PR #116, opened 2021-07-01
More diagrams

oscarlevin:
Add diagrams support for all 5 conversion targets, selected by what is in source.

also fixed a bug/typo introduced in the previous pull request.


oscarlevin:
I've also updated the core commit to see if there are changes to pretextcore we want.  Remember to run the update script.

StevenClontz (approved):

PR #121, opened 2021-07-07
WeBWorK hotfix (0.3.1?)

oscarlevin:
Fix outdated variable for default ww server; add webwork example to book template and publication file (as comments).

Also fixed two minor things in the pretext build click options: updated help for --stringparam and added 'png' and 'all' to list of diagram output formats (now that the windows issue is confirmed to not be widespread).


StevenClontz:
Thanks - I'm going to try this on my end later today. Assuming I can build webwork, I'll bump the version number and push a release to PyPI. (I don't think we need to do an rc)

PR #133, opened 2021-07-29
Print Targets

### LINKED ISSUE #127: `pretext --targets` outputs a list of valid targets from manifest
This is a little bit of a developer request.  For the VS Code pretext-tools, I have currently hard coded in html and pdf as build targets.  But ideally, the targets that the user can select should be those for which there are aliases in the project manifest.  Now I could figure out how node lets you read in and parse the xml file and recreate all the logic, but it would be so easy to use what we already have in the CLI to simply output a list of outputs that I can then parse.

I'm happy to implement as long as @StevenClontz thinks this is an okay idea.

oscarlevin:
Two changes:
- Removed imports that never got used
- Implemented --targets option to print the names of targets in the project manifest (useful for the vs-code extension).

Not sure if I did this is a object-oriented enough way.  There was a method "targets" for Project(), but perhaps that is broken?  (The `xml_element` in the `for` statement is not defined?)

This would close #127.

### DIFF
--- a/pretext/build.py
+++ b/pretext/build.py
@@ -1,11 +1,10 @@
 from lxml import etree as ET
 import logging
 import os
-import shutil
 import sys
 import pathlib
 
-from . import static, utils
+from . import utils
 from .static.pretext import pretext as core
 
 # Get access to logger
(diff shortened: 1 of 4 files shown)


StevenClontz (commented):
I suggest in a comment a way to fix `Project().targets()` which might keep the printout a little more maintainable.

Issue #129, opened 2021-07-22
Wrap all calls to core PreTeXt python script with `utils.working_directory` context manager


StevenClontz:
Right now the core PreTeXt python script will working directory to a temporary location without cleaning up afterward, which means chaining a call to the core script with a call that assumes an unchanged working directory will behave very unexpectedly.

One way to circumvent this is to sequester any calls to the core script with our `working_directory` context manager:

```
from .utils import working_directory
from .static.pretext import pretext as core

with working_directory("."):
    # no matter what happens here, the working directory will return to the original directory when finished
    core.whatever()
```

@rbeezer could alternatively implement the pattern at https://github.com/PreTeXtBook/pretext-cli/blob/main/pretext/utils.py#L18 in the core script to clean up things on that end.

PR #135, opened 2021-07-29
Better placement of build messages

### LINKED ISSUE #130: fix hardcoding of suggesting to `pretext build html`/`latex` in build script
e.g. https://github.com/PreTeXtBook/pretext-cli/blob/main/pretext/build.py#L24

Not every html build has the target name `html`. And some builds are part of `pretext view --watch` where the prompt to view isn't necessary. So that nudge should be in `Project().build` instead, where we know the name of the target.

oscarlevin:
Move success and error messages out of build.py into project.py, to correctly specify view commands.

Also update core commit (to incorporate newest directory management).

Closes #130


StevenClontz:
Added a few tweaks in cfdbdca to fix scenario where target is inferred implicitly and to address #129 - merge and close if you're happy with those.

Issue #101, opened 2021-06-22
Check whether input file is a valid pretext (root) file


oscarlevin:
Related to #100 and #66 , we should check whether the input file is the root of a pretext document and provide suggestions if not.

oscarlevin:
Closed by #143

PR #143, opened 2021-07-30
Better checks

### LINKED ISSUE #100: Check whether publication file is a publication file
See https://groups.google.com/g/pretext-support/c/ei0Dk8YfdPY.

The issue, I think, is that the publication file that is passed isn't actually a publication file of the sort that the xsl expects.  It would be easy to verify that the file passed has the correct format.  At the same time, we should check that the file exists.

oscarlevin:
Check root elements of publication and main source files.  Improve naming of these functions.

Closes #100 and #101 .

### DIFF
--- a/pretext/project.py
+++ b/pretext/project.py
@@ -186,7 +186,7 @@ class Project():
         if not self.xml_publication_is_valid(target_name):
             return
         # Validate xml against schema; continue with warning if invalid:
-        self.xml_source_validate(target_name)
+        self.xml_schema_validate(target_name)
         # Ensure directories for assets and generated assets to avoid errors when building:
         target = self.target(target_name)
         utils.ensure_directory(target.external_dir())
--- a/pretext/project.py
+++ b/pretext/project.py
@@ -25,7 +25,7 @@ class Target():
                 static_dir = os.path.dirname(static.__file__)
                 template_xml = os.path.join(static_dir,"templates","project.ptx")
                 xml_element = ET.parse(template_xml).getroot().find("targets/target")
-                publication = os.path.join(static_dir,"templates","project.ptx")
+                publication = os.path.join(static_dir,"templates","publication.ptx")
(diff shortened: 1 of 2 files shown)


pretext/project.py
```diff
@@ -295,13 +297,28 @@ def publish(self,target_name):
         log.info(f"Latest build successfully pushed to GitHub.")
         log.info("(It may take a few seconds for GitHub Pages to reflect any changes.)")
 
-    def xml_source_syntax_is_valid(self,target_name):
+    def xml_source_is_valid(self,target_name):
         target = self.target(target_name)
         return utils.xml_syntax_is_valid(target.source())
 
-    def xml_source_schema_is_valid(self,target_name):
+    def xml_source_validate(self,target_name):
```
StevenClontz, line 304: Maybe `xml_schema_validate` or `xml_schema_is_valid`? I'll never remember the difference between `_is_valid` and `_validate` suffixes.

oscarlevin:
```suggestion
    def xml_schema_validate(self,target_name):
```

pretext/project.py
```diff
@@ -181,10 +181,12 @@ def build(self,target_name,webwork=False,diagrams=False,diagrams_format="svg",on
...
         self.init_ptxcore()
         # Check for xml syntax errors and quit if xml invalid:
-        if not self.xml_source_syntax_is_valid(target_name):
+        if not self.xml_source_is_valid(target_name):
+            return
+        if not self.xml_publication_is_valid(target_name):
             return
         # Validate xml against schema; continue with warning if invalid:
-        self.xml_source_schema_is_valid(target_name)
+        self.xml_source_validate(target_name)
```
oscarlevin, line 189:
```suggestion
        self.xml_schema_validate(target_name)
```

StevenClontz (approved):

PR #161, opened 2021-08-05
Allow for custom xsl

### LINKED ISSUE #23: Call arbitrary xsl file
`pretext build custom`

### LINKED ISSUE #160: Signature changes for PDF, HTML, LaTeX conversions in core
The `pdf()`, `latex()`, and `html()` routines now have a new fourth argument, which is a stylesheet that will be used instead of the default versions.

You can pass `None` in this parameter until you want to support it.  To use, it should receive a full filename (complete path).

It is the author's responsibility to ensure it behaves!  Which could suggest a reason for `pretext/pretext` to have this support, and the CLI not to?

See https://github.com/rbeezer/mathbook/commit/5f6cdebcf0377835aec7e503010e4ea3ee0cd40a

oscarlevin:
Now a user can specify `<xsl>xsl/custom-html.ptx</xsl>` in a target in the project manifest.  That file can import using `@pretext-href="pretext-html.xsl"` and it will import the static xsl correctly.

Currently, no command line option for this, but that would be possible if we thought it was important.

This closes #23  and closes #160.

If we are happy here, I think it warrants releasing 0.5


oscarlevin:
I went ahead and added the command-line override since it was easy (I think, if I did it right).

StevenClontz:
Sorry for the delay here - kid's daycare is out yesterday/today (planned,
although kind of grateful with the delta surge happening here). I plan to
catch up this weekend.


Steven Clontz

StevenClontz:
working on a 0.5rc0 release. I think I want to change up our versioning workflow so that `main` is +0.0.1 the previous release, and if we decide that it deserves rounding up to the next 0.1 then we can make that change.

oscarlevin:
Sounds good.  Thanks for doing that.

StevenClontz:
https://pypi.org/project/pretextbook/0.5.0rc0/

StevenClontz:
Wire that up with a production book and let me know if it's ready for a public release

oscarlevin:
Good thing I checked.  There is an issue on Windows with the `xsl:import` URI that the script inserts.  I'm going to try to to pass the linux path thing and see if that helps.

On WSL, it works beautifully.

PR #163, opened 2021-08-08
fix path URI issue with xsl:include on Windows

oscarlevin:
I think this takes care of it, but for some reason I can't pip install from the remote for some reason (`UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1663: character maps to <undefined>`).  Can you build RC1 after accepting this and I can test again, please?

### DIFF
--- a/pretext/build.py
+++ b/pretext/build.py
@@ -2,8 +2,6 @@ from re import T
 from lxml import etree as ET
 import logging
 import os
-import sys
-import pathlib
 
 from . import utils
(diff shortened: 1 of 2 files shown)


oscarlevin:
This was successful in both windows and wsl.  So I'm ready of 0.5.

StevenClontz:
https://pypi.org/project/pretextbook/0.5.0/


Steven Clontz

PR #173, opened 2021-10-29
Update core

### LINKED ISSUE #168: Don't copy .xsl or .ent files to output
When doing a custom xsl build, we copy all the user's `xsl` directory to the temporary directory (replacing pretext-href with the appropriate href).  However, when the build finishes, the last step is to copy the output that was placed in the temporary directory into the output folder for the project.  Currently, all the .xsl and .ent files are being copied as well.  This doesn't break anything, but is sloppy and annoying.

### LINKED ISSUE #170: match `error-syntax` in gitignore with `error_syntax` files
At some points hyphens were changed to underscores and vice versa - we need to ignore those build error files.

oscarlevin:
In addition to updating the core commit (tested), this pull request fixes two minor bugs: 

- hyphens are replaced with underscores in the templates gitignore files, closes #170
- Puts any custom xsl in its own temporary directory so that it doesn't get copied to the output folder.  Closes #168 

If this looks good, can you @StevenClontz push the new version to PyPi?

(sorry if this is based to the wrong branch; I don't remember how we are supposed to branch bugfixes)


StevenClontz:
I think we changed development to always be against main, to make PRs easier. (-:

StevenClontz:
Please install https://pypi.org/project/pretextbook/0.5.2rc0/ and see if it's working as expected.

oscarlevin:
Looks like the RC is working correctly.  Should be good to push a regular release.

StevenClontz:
Well, somewhere between last week and this week my CoCalc project started refusing to `update_core` from GitHub, so I need some time to fix that before I can push the public release...

StevenClontz:
Okay fixed, 0.5.2 is live https://pypi.org/project/pretextbook/0.5.2/

PR #174, opened 2021-11-05
fix bug when no custom xsl is copied

oscarlevin:
In the last update, I introduced a new bug that caused an ugly error to show up when custom xsl is not used.  This small fix takes care of that.  I suggest we push this immediately as a bugfix.

### DIFF
--- a/pretext/project.py
+++ b/pretext/project.py
@@ -226,6 +226,7 @@ class Project():
                 shutil.rmtree(target.output_dir())
         #if custom xsl, copy it into a temporary directory (different from the building temporary directory)
         custom_xsl = None
+        temp_xsl_dir = None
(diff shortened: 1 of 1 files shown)


StevenClontz:
Of course, this is why we should have a test suite, at least an automated build of several sample projects to ensure no regressions cause things to stop building that used to build.

StevenClontz:
try https://pypi.org/project/pretextbook/0.5.3rc0/

StevenClontz:
actually given the nature of the error (custom XSL related, and the default pretext new build seems to work fine with this change), I'm going ahead and pushing a public release.

StevenClontz:
Now live at https://pypi.org/project/pretextbook/0.5.3/

Issue #213, opened 2022-03-11
Document custom XSL support


oscarlevin:
For folks developing pretext-core, it would be helpful to expose the most recent commits from pretext via the cli.  This could be accomplished by having an argument that specifies a core-pretext directory (that the user git-pulls from the pretext project on github).

StevenClontz:
I think https://github.com/PreTeXtBook/pretext-cli/blob/main/scripts/symlink_core.py does much of this. What else do we want to add to it?

StevenClontz:
Oh wait I see; this is so folks would only need to *install* the CLI, not have a copy of an in-development version.

So begrudingly, I think we *do* need to be installing core XSL to `~/.pretext/xsl` in order to solve #180 anyway. So it's not a stretch to add a sibling to `<targets/>` called e.g. `<core-xsl-path/>` to point to a different copy of core XSL.

What we *cannot* do is support swapping out core python. If you want to mess with core Python and the CLI then you need the in-development CLI and need to use symlink_core.py.

oscarlevin:
We now have the ability to use core pretext's xsl via #217.  You can simply specify `<xsl>path/to/pretext/xsl/pretext-html.xsl</xsl>` or the like.  

Do we want to be more explicit about this?  Document it somewhere other than here?  Have a child of `<target>` that says to build against a local core pretext, but still use a separate custom xsl?

StevenClontz:
We should definitely include this as a commented option on our template project.ptx.

StevenClontz:
@oscarlevin put your eyes on 41c3f45 and let me know if it works

oscarlevin:
I added a comment on the commit about one of the comments.  I think this is helpful, although we will want some proper documentation eventually.

StevenClontz:
With your approval of the PR, I hope to soon package up 0.6.2 and submit a PR to core PreTeXt documentation updating the "getting started" section to include deployment to GitHub Pages

StevenClontz:
I definitely don't like `pretext new` building a folder for custom XSL - that seems to send the wrong message. So I think I like removing the commented-out `custom` target from generated project.ptx altogether, and putting this info in documentation only.

StevenClontz:
Removing this from a `pretext new book`

```
    <!-- Custom XSL may invoke core PreTeXt by using pretext-href -->
    <!-- Only files in the directory containing custom.xsl are used -->
    <!-- <target name="custom">
      <format>html</format>
[...]

Issue #201, opened 2022-03-04
suggest deprecating os file stuff in favour of pathlib.


arechnitzer:
What it says in the title.

(but is only a suggestion)

rbeezer:
I tried to use pathlib in the core routines, in order to get Windows compatibility, and it was a disaster.

arechnitzer:
Strange... I thought the main point of pathlib was to avoid "I want my code to hack files on linux/osx/windows"-problems

rbeezer:
Exactly.  Could be an ID-10T problem, but I gave it a serious go.  No love.

I think there may be just one place where we end up sanitizing path separators.  Where a publication file comes in with an OS-specific filename, and then needs to opened/read by XSL. Maybe.

arechnitzer:
Ah... "fun".

PR #217, opened 2022-03-12
Copy all files/subdirs that are peers of custom xsl

siefkenj:
This PR offers a partial solution to https://github.com/PreTeXtBook/pretext-cli/issues/213

At the moment, all peer `xsl` from a custom-specified xsl file is copied. However, no subdirectories or `xml` is copied. That means if you specify, for example, `pretext-html.xsl` from git master as your custom `xsl`, the `translations` subdirectory will not be copied. The result is many compile errors!

This PR
  1. Copies all peer subdirectories and files to a custom `xsl`. The files are not limited to the `xsl` extension
  2. Renames some references to `rbeezer/mathbook` to reference `pretext` (the current project name).

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,14 +193,14 @@ need to be rebuilt by each user when pulled from GitHub.
 
 The file `pretext/static/CORE_COMMIT` tracks the upstream
 commit of core PreTeXt XSL/Python code we're developing against
-(from `rbeezer/pretext`).
+(from `PreTeXtBook/pretext`).
 To grab these updates from upstream, run:
 
 ```
 python scripts/update_core.py
 ```
 
-If you instead want to point to a local copy of `rbeezer/pretext`,
+If you instead want to point to a local copy of `PreTeXtBook/pretext`,
 try this instead to set up symlinks:
 
 ```
--- a/README.md
+++ b/README.md
@@ -193,18 +193,18 @@ need to be rebuilt by each user when pulled from GitHub.
 
 The file `pretext/static/CORE_COMMIT` tracks the upstream
 commit of core PreTeXt XSL/Python code we're developing against
-(from `rbeezer/mathbook`).
+(from `rbeezer/pretext`).
 To grab these updates from upstream, run:
 
 ```
 python scripts/update_core.py
 ```
 
-If you instead want to point to a local copy of `rbeezer/mathbook`,
+If you instead want to point to a local copy of `rbeezer/pretext`,
 try this instead to set up symlinks:
 
 ```
-python scripts/simlink_core.py path/to/mathbook
+python scripts/symlink_core.py path/to/pretext
 ```
 
 Updates to `templates/` must be zipped and moved into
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -193,18 +193,18 @@ need to be rebuilt by each user when pulled from GitHub.
 
 The file `pretext/static/CORE_COMMIT` tracks the upstream
 commit of core PreTeXt XSL/Python code we're developing against
-(from `rbeezer/mathbook`).
+(from `rbeezer/pretext`).
```
StevenClontz, line 196: should actually be `PreTeXtBook/pretext`

scripts/symlink_core.py
```diff
@@ -15,4 +15,4 @@ def remove(path):
     remove(link_path)
     os.symlink(original_path,link_path)
 
-print("Linked local rbeezer/mathbook ")
+print("Linked local pretext directory")
```
StevenClontz, line 18: I'd slightly prefer "core pretext", what do you think?

pretext/utils.py
```diff
@@ -234,16 +235,27 @@ def expand_pretext_href(lxml_element):
     for ele in lxml_element.xpath('//*[@pretext-href]'):
         ele.set('href',str(linux_path(static.core_xsl(ele.get('pretext-href'),as_path=True))))
 
-def copy_fix_xsl(xsl_path, output_dir):
```
StevenClontz, line 237: @oscarlevin this is your function I think - can you review this?

StevenClontz:
Thanks! I think this is mostly Oscar's concern so I've asked him to review.

scripts/symlink_core.py
oscarlevin: As this is just for developers, I agree, "Linked local core pretext directory" is clearer.

pretext/utils.py
```diff
@@ -234,16 +235,27 @@ def expand_pretext_href(lxml_element):
...
     with working_directory(xsl_dir):
-        for filename in os.listdir('.'):
+        for filename in glob.iglob('**', recursive=True):
+            output_path = os.path.join(output_dir, filename)
+            if not os.path.isfile(filename):
+                # glob lists both files and directories, but we only want to copy files.
+                continue
+            # Ensure the existence of the subdirectory our file is in before writing.
+            pathlib.Path(os.path.dirname(output_path)).mkdir(parents=True, exist_ok=True)
+
```
oscarlevin, line 253: Do we use pathlib anywhere else?  I would prefer we are consistent with our file management methods.

oscarlevin:
I think this does what we want, but I want to test it and see if we can use the same sorts of file management calls that we use elsewhere.

pretext/utils.py
siefkenj: `linux_path` already used it which is why I used it again.

StevenClontz: See also #201 which I think is a good idea anyway

StevenClontz:
I'm happy when you're happy Oscar (or let me know if you need me to test things out).

oscarlevin:
Sorry for the delay.  This is a nice addition and essentially solves #213, although we might want to document this as the standard way to accomplish this.

PR #244, opened 2022-06-14
set major versions in setup.py

StevenClontz:


StevenClontz:
Brad suggested in Zoom today that setting major versions is a good idea. Before I merge I'll make sure that a pip install works, and also take the time to remove any possible unused dependencies (I don't think we're using slugify anymore).

PR #246, opened 2022-06-14
improve pretext init 

### LINKED ISSUE #95: `pretext init` builds `project.ptx` and `requirements.txt` and `publication/publication.ptx`

### LINKED ISSUE #241: `pretext init` gitignore file
See discussion in https://github.com/PreTeXtBook/pretext-cli/issues/195 - to prevent `pretext deploy` from adding any nasty build files inadvertently, we should provide a template `.gitignore` when initializing an existing project.

oscarlevin:
Add timestamped versions of files that exist; add force option.  Added very bare bones .gitignore file that is copied (will update soon).  Closes #95 and closes #241

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -2,6 +2,7 @@ import click
 import click_logging
 import logging
 import shutil
+import datetime
 import os, zipfile, requests, io
 import tempfile, shutil
 import platform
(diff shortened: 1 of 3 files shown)


oscarlevin:
Upon further reflection, I'm unsure whether we are doing this the right way.  When a user already has at least some of the files `init` provides, the current proposal copies the "standard" files to a timestamped version, leaving the files already in the project alone.  If the user already has a `project.ptx` file, nothing is done, unless the user types `pretext init --force`.  In this case, everything, including `project.ptx` gets a new file, possibly timestamped if the file already exists in the expected place.

But if we are really "forcing" the init, then perhaps we should copy the current files to timestamped backups (say `project-bk-date.ptx` for example).  Thoughts?

oscarlevin:
Getting the new boilerplate is exactly the use case I was trying to support.  I guess I slightly prefer `--force` over `--update` since this is not something people will want to do often (like upgrading packages might), and both have the same issue as sounding like they replace the current setup (which I am happy to not do, in favor of the currently proposed timestamped new files).

If this is satisfactory for now, I'll let you merge the pull request and we can always rename the option later.

StevenClontz:
I toyed with this a little but didn't come up with any better behavior, and trashed my work.

My problem is that generally `--force` commands are bad ideas unless you're an expert; I think we'll want novice authors to occassionally update their CLI and let our `init` subcommand facilitate this update.

But merging for now.

PR #251, opened 2022-06-16
Build generate assets

### LINKED ISSUE #245: Consider changing `--diagrams|-d` to `--generate-assets|-g`
Since many of the steps we're doing in `--diagrams` aren't diagrams, maybe this should be renamed.

StevenClontz:
Closes #245 

Cleans up build options and adds a separate generate subcommand to handle only generating assets without building the rest of the project.


oscarlevin (approved):
Very nice.  Haven't tested at all, but the structure is excellent.

PR #255, opened 2022-06-17
Change target names to not match output format

StevenClontz:


StevenClontz:
With no objections, I'm going to merge this. I think there's a lot of value in distinguishing the *target name* from the *output format* as we educate authors on the CLI next week.

Issue #286, opened 2022-06-30
Epub/Kindle/Braille issues


oscarlevin:
As in #30 , we would like to support building epub, kindle, and braille with the CLI.  Here are the current stumbling blocks:

1. We would need to ship the `script\mjsre` folder from core pretext which has a js script `mj-sre-page.js` that gets called.  This in turn requires a few packages that should be installed into the `mjsre` folder (a global `npm` install of these packages doesn't seem to work.  This means we either need to test installing the entire scripts folder into the user's home directory (say in `.pretext/script/mjsre`) and have them npm-install the packages (and this might not even work, depending on what assumptions `pretext.py` makes about the paths to that folder), or include the entire node_modules folders that get installed when the requirements are installed.
2. Likely we also need core's `CSS` folder, since the script copies local css to do the formatting.  This is less of a big deal.
3. I have not been able to test this out, because on Windows I get a file permission error.  I think I get this no matter what (using `pretext/pretext` for example), so it seems like even if we get these supported, we can't support windows as is.

So the question is, do we proceed?

rbeezer:
Braille requires `liblouis` which I install from source.  Usually OK, but sometimes some gotchas.  If you didn't do braille, you don't need SRE (modulo the "speech enhancement" we uncovered in Ann Arbor, which maybe should move under some flag/switch).  That leaves MathJax for EPUB, no way around that.  At least MJ has a broad base of users who try to use it locally (or server-side).

@mitchkeller is the one to speak to Kindle.

mitchkeller:
1. I don't think it would be awful if the CLI shipped a node_modules folder that had the versions that the current `package.json` file specifies. The publisher would still need to have `node` installed, but that's no different than requiring that they have a TeX distribution installed to build PDF.
2. Yes, this folder is going to be required.
3. ¯\_(ツ)_/¯ 

I think once epub-svg builds, epub-kindle should mostly be fine. Well, that's assuming the routines that build generated assets can produce png files for those images. Non-generated assets will also need to have a png version available (for things that are SVG in HTML), but I feel like it's already something we assume that authors do for themselves.

oscarlevin:
The windows permission error should be fixed by https://github.com/PreTeXtBook/pretext/pull/1779

oscarlevin:
The issues here are resolved, even though braille is still in the works and there are a few tweaks needed for epub/kindle.  All tracked in #30

PR #290, opened 2022-07-10
set up local core install

### LINKED ISSUE #180: Install PreTeXt resources to user directory
The more I learn about how Python/PyPI distributes files ("resources") in packages, the more I worry about our brittle handling of static files. I think the more Pythonic solution is to ship a ZIP of the XSL files, install them to `~/.pretext/` or similar, and use them from there.

The problem with our current solution is that we try to grab a filepath to XSL assuming resources exist on disk, but it's feasible that a user might install the package as a zip, so the file doesn't actually exist. XSL cannot work together unless they actually live on the disk in the expected relative paths, so we need to put them on disk somewhere.

oscarlevin:
With this commit, anytime `pretext` is run, we check whether the local version of the static files exist in `~/.ptx` (well, actually we just check for a `.commit` file with the commit corresponding to `CORE_COMMIT`).  If not, the (new) files are installed.  

Templates are still stored in the previous spot; so the call to that location had to be updated (now `static.templates_path()`.  

The `update_core.py` script now creates a zip of the core files in folders `css`, `schema`, `script`, and `xsl`.

This is a required change to implement #286 and will close #180

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -139,6 +139,7 @@ pretext/static/pretext
 pretext/core/pretext.py
 pretext/static/schema
 pretext/static/xsl
+pretext/static/static.zip
 #shipped templates
 pretext/static/templates
 
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -7,7 +7,6 @@ import os, zipfile, requests, io
 import tempfile, shutil
 import platform
 from pathlib import Path
-import sys
 from typing import Optional
 
 from . import utils, static, VERSION, CORE_COMMIT, core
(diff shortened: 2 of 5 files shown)


oscarlevin:
Apparently this broke the tests.  I tested locally, perhaps the tests are broken by this change.  I will investigate after lunch.

oscarlevin:
I missed a call to an unnecessary `core_xsl` function.  Fixed in the second commit.  Tests now pass.  @StevenClontz , go ahead and merge if you are comfortable with the structure I employed.

PR #297, opened 2022-07-18
Improve pretext view, especially for non html formats

oscarlevin:
I've attempted to use the `webbrowser` module to automatically open the localhost when we run `pretext view` and to open the other file when it is not an html formatted target.  I have no idea what happens with this in cocalc (I tried it, but don't have internet access, so perhaps that's why nothing happened).  If pdf doesn't do something nice in cocalc, we could serve that like html for cocalc users.

What do we want to do about view for non-html targets?  These seem like they would always take longer to build (I guess latex is pretty quick, but there is no reason you would want to get that output while watching).  We can just warn the user that the watch option only works with html builds, or try to do something more clever.

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -385,7 +385,7 @@ def generate(assets:str, target:Optional[str], all_formats:bool, xmlid:Optional[
     type=click.Choice(ASSETS, case_sensitive=False), 
     help='Generate all or specific assets before viewing')
 @click.option(
-    '--no_launch', is_flag=True,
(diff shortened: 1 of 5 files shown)


oscarlevin:
Address #299 before merging

oscarlevin:
Also take care of better window opening for cocalc, as mentioned in #298

StevenClontz:
oops 5f105db was a stab in the dark. Going to make a proper fix now...

PR #302, opened 2022-07-18
factor out cocalc logic in utils

StevenClontz:


StevenClontz:
Closes https://github.com/PreTeXtBook/pretext-cli/issues/298

PR #307, opened 2022-07-19
remove 'static' in favor of core.resources and templates modules

StevenClontz:
work in progress - tests should be passing but need to fix `symlink_core`


StevenClontz:
@oscarlevin observe the changes: `update_core.py` to `fetch_core.py`, all things we used to call "static" are now called "resources" to be a bit more Pythonic, and we have a `core/resources.zip` for stuff from PreTeXtBook/pretext and a `templates/resources` for the stuff we're zipping up into our package.

Tests are passing but once this is in main I'm gonna push an alpha release to make sure the package is in fact building correctly.

Issue #283, opened 2022-06-29
pyppeteer dependency


StevenClontz:
Add pyppeteer dependency, make sure interactive screenshots work.

StevenClontz:
Moving to 0.9 - for 0.8 I've simply disabled interactive previews in 0de0f6e - they aren't needed for HTML builds, and their absence doesn't break LaTeX/PDF builds. Right now core asset generation assumes the HTML has already been deployed at the location configured by https://pretextbook.org/doc/guide/html/publication-file-online.html#online-baseurl but this needs to be fixed such that HTML need not have ever been built, much less deployed to the public web.

Issue #317, opened 2022-07-26
add codechat dependency


StevenClontz:
https://codechat-system.readthedocs.io/en/latest/CodeChat_Server/install.html#id3

StevenClontz:
N.B. this was reverted in #339 to reduce the time for installation.

PR #323, opened 2022-07-30
add codechat and pyppeteer dependencies, get interactive screenshots working

StevenClontz:


StevenClontz:
Closes #283 and #317, but I was only able to mock up a skipped test for interactive preview generation since upstream needs updates to build these previews locally.

PR #339, opened 2022-08-14
remove codechat (and 23 other subsequent) dependencies

StevenClontz:


StevenClontz:
Just pushed 1.0.0 and realized I forgot this offline decision to remove the hefty codechat dependency. I'll release 1.0.1 to remove them (folks can separately `pip install codechat-server` if they want still).

PR #360, opened 2022-11-03
edits to webwork generation to support/fix xmlid_root

StevenClontz:


StevenClontz:
Silly me took a while to remember that while nightlies pull latest core commits, when we change the CLI we need to manually redirect that hash to grab the updates...

oscarlevin (approved):
Looks good

PR #366, opened 2022-12-09
Reformat with pep8

siefkenj:
Reformat the python codebase according to `pep8` rules. Using `ctrl+shift+I` in vscode should perform the exact same formatting that is in this PR. Subsequent applications of autoformatting shouldn't change anything.

This PR touches a lot and might be annoying to merge if there is current work on many branches, but hopefully that's not the case!

This PR also fixes some spelling errors, but makes no other changes.

### DIFF
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -44,3 +44,7 @@ jobs:
     - name: Test with pytest
       run: |
         python -m poetry run pytest -v
+
+    - name: Check formatting with black
+      run: |
+        python -m poetry run black --check $(git ls-files |grep -e "\.py$")
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -15,7 +15,7 @@ jobs:
       with:
         python-version: "3.10"
 
-    - name: Install poetry ${{ matrix.poetry-version }}
+    - name: Install poetry 1.1.13
       run: |
         python -m ensurepip
         python -m pip install --upgrade pip
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -13,7 +13,7 @@ jobs:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
       with:
-        python-version: 3.10
+        python-version: "3.10"
 
     - name: Install poetry ${{ matrix.poetry-version }}
       run: |
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -7,6 +7,29 @@ on:
     branches: [ main ]
 
 jobs:
+  format:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      uses: actions/setup-python@v4
+      with:
+        python-version: 3.10
+
+    - name: Install poetry ${{ matrix.poetry-version }}
+      run: |
+        python -m ensurepip
+        python -m pip install --upgrade pip
+        python -m pip install poetry==1.1.13
+
+    - name: Install dependencies
+      shell: bash
+      run:  |
+        python -m poetry install
+
+    - name: Check formatting with black
+      run: |
+        python -m poetry run black --check --diff $(git ls-files |grep -e "\.py$")
+
   build:
     strategy:
       fail-fast: false
@@ -44,7 +67,3 @@ jobs:
     - name: Test with pytest
       run: |
         python -m poetry run pytest -v
-
-    - name: Check formatting with black
-      run: |
-        python -m poetry run black --check $(git ls-files |grep -e "\.py$")
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -28,7 +28,7 @@ jobs:
 
     - name: Check formatting with black
       run: |
-        python -m poetry run black --check --diff $(git ls-files |grep -e "\.py$")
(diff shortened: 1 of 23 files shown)


bnmnetp:
ctrl-shift-I can mean many things depending on how you have your formatter configured.  It may use black or it may use pep8 or something else.

I would suggest we use black as it is widely used in the Python community and easy to add to CI as a part of evaluating a new PR.

Also massive reformats like this should be undertaken with care as it does essentially wipe out the history of who wrote what replacing it with who ran the reformatter and now looks like they wrote everything.  There are ways to help git blame get past this, but require a bit of work.

siefkenj:
`black` is fine as long as it's something :-). `autopep8` seems to be the default that is used in vscode, so that's my personal preference. We can add a formatting check in the CI if we want to ensure everything is properly formatted in the future.

It's a disruptive change, but like I said, I think it is worth it.

@StevenClontz @oscarlevin  Do you have thoughts?

StevenClontz:
I can't find the PR but I thought @arechnitzer had suggested this a while back. If so don't remember how that played out.

I'm anticipating suggested contribution/development workflow will soon we via Codespaces we can bootstrap however we want. And I like sharing a Python style guide across all PreTeXt-Runestone projects.

So I'll happily +1 a contibution that sets this up nicely, with documentation in the README to get this set up manually (we can figure out Codespace stuff later).

arechnitzer:
I strongly recommend using python-black to facilitate this... and maybe flake8 too. Saves much pain down the line. 

*but* when you do blacken/flake8 your code base it will look like a HUGE change since everything gets reformatted.

arechnitzer:
Another benefit of black / flake8 is that they work anywhere and dont need vscode. You can also work them into your CI so that a pull-request is rejected unless the code has been formatted "correctly". 

Now - black is not perfect, you might not like all its formatting choices, but it definitely has the benefit of being the same everywhere.... and generally makes your code legible.

siefkenj:
I have updated formatting to use `black` and added a CI test to enforce proper formatting in the future. A line was also added to the `README.md` explaining how to format all files.

Hopefully this can get merged soon, my next PR will look massive until this is in :-)

oscarlevin:
I don't have strong opinions about this, but I like the idea of having a standard.  @StevenClontz , I'll leave the decision to you.

bnmnetp:
If we want to be consistent then I would also vote for `black`. It has served us well for that past few years.

rbeezer:
> Also massive reformats like this should be undertaken with care as it does essentially wipe out the history

And you know, I love my git history.  ;-)

> You can also work them into your CI so that a pull-request is rejected unless the code has been formatted "correctly".

There's a way to welcome a new contributor!  ;-)

When somebody new makes a PR without proper formatting, do you:

* Tell them to resubmit properly?
* Reformat their work for them?
* Accept it as-is, and do a massive automated clean-up later?

I incrementally clean-up my XSL when I refactor, or am in the neighborhhod.  Because, you know, I love my git history.  And also want to encourage new contributors

siefkenj:
I think if a user makes a failed PR and is asked:

Please run `poetry run black --check --diff $(git ls-files "*.py")` in the terminal and push one more time.

I think it won't be overly discouraging (especially if they went through all the pain of getting the dev environment set up, which, I can attest, is still not so easy...)

StevenClontz:
Clarification: as I recall the PR isn't "rejected", it's just marked as "not ready to be merged".

Is there a way to simply apply the codestyle automatically by an Action?

Also, the preferred dev workflow I want for casual contributors will likely be Codespace-based in the coming weeks - we can make sure that VS Code is set up to Just Work(TM).

siefkenj:
@StevenClontz I am not sure of a clean way to do it with an Action...One idea is that you could have a "staging" branch. Code is **only** ever merged into the staging branch. The staging branch would then format and make a PR to main. (This doesn't sound like a good idea to me...)

You can set up a git pre-commit hook to format files. That works for some people.

What should be not too hard (though I don't know how to do it) is to set up an action to automatically add a comment with appropriate instructions to a PR whose `format` check fails.

StevenClontz:
Before I squash/merge, I want to make sure I understand [this](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html) correctly: by creating a `.git-blame-ignore-revs` file with the commit from this merge, that should be all we need to prevent `git blame` inaccuracies despite the many lines touched here?

Re @siefkenj thanks for the feedback. I think we can see how this goes and decide how exactly to streamline contribution workflows down the road a bit.

bnmnetp:
Yeah, Reject is definitely too strong.

We just have an action that runs that includes:

1. Run black to check formatting
2. Run flake8 for lint type Python problems
3. Run mypy to do some static typechecking - at least on bookserver where we have been diligent about type hints.
4. Run unit tests

In theory, this workflow must run successfully before we merge a PR.

My modus operandi has always been that a PR is the start of a conversation not a perfectly formed commit ready to merge.

siefkenj:
@StevenClontz It looks like a nice workflow might be possible: https://mskelton.medium.com/auto-formatting-code-using-prettier-and-github-actions-ed458f58b7df  Though I suggest merging this first and enhancing the workflow in a followup PR.

arechnitzer:
Why not set up the CI so that "when you push unblackened code to any branch except main you get a warning. If you push unblackened code to main then it gets rejected" ?

StevenClontz:
Looks good on GitHub: https://github.com/PreTeXtBook/pretext-cli/blame/main/pretext/__init__.py

StevenClontz:
Trying out formatting CI check at #369

Issue #327, opened 2022-08-01
Support programmatic creation of a project


StevenClontz:
Right now a new `Project()` may only be instantiated by reading a `project.ptx` off the disk. This is a pain for me today, because (for reasons) I have a project that needs two separate targets for each section with custom XSL. So now I'm creating this `project.ptx` by hand, but I'd have preferred to programmatically instantiate a `p = Project(xml_object)`.

oscarlevin:
This is the major feature of the 2.0 release.  @StevenClontz, you should write something for the guide about how use this feature.

PR #367, opened 2022-12-10
Add ability to overlay project.ptx from the commandline

siefkenj:
As discussed on the mailing list, this PR adds the ability to overlay settings in `project.ptx` from the command line.

For example, if your `project.ptx` file didn't have custom xsl or you wanted to override the custom xsl file, you could run

```
pretext build -P targets.target.xsl "my/custom/path.xsl"
```

If you wanted to change an attribute, you can use the `@` syntax.
```
pretext build -P targets.target@pdf-method pdflatex
```

Multiple instances of `-P`/`--project-ptx-override` can be applied. The largest limitation is that it is based on a simple selector language, using `.` to indicate node hierarchy. There is no way to distinguish between siblings with the same name, so if your tree is `<a><b>hi</b><b>there</b></a>`, using `-P a.b foo` will result in `<a><b>foo</b><b>foo</b></a>`. I think this is a reasonable limitation. If you want full control over the `project.ptx` file, you should create one matching your specifications, rather than using a complex query language on the command line.

The benefit of this feature is that it lets you test out small modifications to `project.ptx` without any source modifications. E.g., if you are working in a git-tracked repo and you wanted to test a different output format but didn't want to worry about accidentally committing a changed `project.ptx` file, this would let you specify `-P targets.target.format ...` on the command line :-).

This may be a partial solution for #327 
This PR depends on #366

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -343,11 +343,11 @@ ASSETS = [
     help="Generates assets for target.  -g [asset] will generate the specific assets given.",
 )
 @click.option(
-    "-P",
+    "-p",
     "--project-ptx-override",
     type=(str, str),
     multiple=True,
-    help=xml_overlay.USAGE_DESCRIPTION.format("-P"),
+    help=xml_overlay.USAGE_DESCRIPTION.format("-p"),
 )
 def build(target, clean, generate, project_ptx_override: t.Tuple[str, str]):
     """
@@ -425,11 +425,11 @@ def build(target, clean, generate, project_ptx_override: t.Tuple[str, str]):
     help="Generate all possible asset formats rather than just the defaults for the specified target.",
 )
 @click.option(
-    "-P",
+    "-p",
     "--project-ptx-override",
     type=(str, str),
     multiple=True,
-    help=xml_overlay.USAGE_DESCRIPTION.format("-P"),
+    help=xml_overlay.USAGE_DESCRIPTION.format("-p"),
 )
 def generate(
     assets: str,
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -1,12 +1,18 @@
 from concurrent.futures.process import _threads_wakeups
 from pickle import FALSE
-import logging, logging.handlers, sys
+import logging
+import logging.handlers
+import sys
 import click
 import click_log
 import shutil
 import datetime
-import os, zipfile, requests, io
-import tempfile, shutil
+import os
+import zipfile
+import requests
+import io
+import tempfile
+import shutil
 import platform
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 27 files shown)


oscarlevin:
This will be a nice addition, and help quite a bit with getting the CodeChat working (it can call pretext build web with a custom `html-with-mapping` format).

Should we also make this option available for `pretext generate`?

siefkenj:
@oscarlevin Good point! I have added the functionality to `pretext generate`.

StevenClontz:
Is there a non-manual way to deal with the merge conflicts?

siefkenj:
@StevenClontz Not that I know of...but I took care of all the conflicts. Should be reviewable now!

pretext/cli.py
```diff
@@ -341,7 +342,14 @@ def init(refresh):
     type=click.Choice(ASSETS, case_sensitive=False),
     help="Generates assets for target.  -g [asset] will generate the specific assets given.",
 )
-def build(target, clean, generate):
+@click.option(
+    "-P",
```
StevenClontz, line 346: CLI convention is to use lowercase flags

pretext/cli.py
```diff
@@ -407,8 +424,19 @@ def build(target, clean, generate):
     default=False,
     help="Generate all possible asset formats rather than just the defaults for the specified target.",
 )
+@click.option(
+    "-P",
```
StevenClontz, line 428: `-p`

StevenClontz (changes requested):

pretext/cli.py
siefkenj: Sure, I'll switch it. I put it as a capital because I thought it was an "advanced" option, that most users wouldn't use, so we could save lower-case letters for common options.

pretext/config/xml_overlay.py
```diff
@@ -0,0 +1,151 @@
+import click
+import typing as t
+from lxml import etree as ET
+
+
+def remove_single_newlines(string: str) -> str:
```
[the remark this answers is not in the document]
siefkenj: The trouble is that new lines get carried over when `--help` is used. I didn't want them to mess up the docs, but I also wanted the doc string to be easily editable. I'll move to utils. Maybe I should change the name to `format_for_click_help`?

siefkenj:
Very weird...I don't know why the CI isn't running on my latest commit. Does that make sense to you @StevenClontz ?

rbeezer:
I can't fetch a PR right now.  Maybe GH is borked...

StevenClontz:
Too lazy to do UTC conversion but it seems likely there was an issue now
resolved: https://www.githubstatus.com/history


Steven Clontz

StevenClontz:
Going ahead and approving assuming tests pass. Not sure the reasoning for a `config` module for this to live in but don't want to hold this up further.

StevenClontz (approved):

PR #369, opened 2022-12-12
add un-blackened line

StevenClontz:


pretext/utils.py
```diff
@@ -1,5 +1,5 @@
 from asyncio import subprocess
-import os
+import os, random_package
```
StevenClontz, line 2: I guess this is valid with Black? The PR fails for other reasons (package DNE) obviously but I expected this to get flagged.

PR #370, opened 2022-12-13
Add: Move mapping code which supports the CodeChat System 

bjones1:
See https://github.com/PreTeXtBook/pretext/pull/1867 for additional discussion.

### DIFF
--- a/pretext/build.py
+++ b/pretext/build.py
@@ -4,7 +4,7 @@ from pathlib import Path
 import sys
 from typing import Optional
 
-from . import utils, core
+from . import utils, core, codechat
 
 # Get access to logger
 log = logging.getLogger("ptxlogger")
@@ -39,6 +39,9 @@ def html(
                 custom_xsl and custom_xsl.as_posix(),  # pass None or posix string
                 None,
                 output.as_posix(),
+            )
+            codechat.map_path_to_xml_id(
+                ptxfile, utils.project_path(), output.as_posix()
             )
         except Exception as e:
             log.critical(e)
--- a/pretext/codechat.py
+++ b/pretext/codechat.py
@@ -1,0 +1,110 @@
+# ***********************************************************
+# codechat.py -- Compute mapping used by the CodeChat System
+# ***********************************************************
+# This function computes a mapping between source file names and the resulting HTML files built from them by PreTeXt. The `CodeChat System <https://codechat-system.readthedocs.io>`_ then uses this mapping to synchronize between a source file and its resulting rendered form.
+#
+#
+# Imports
+# =======
+# These are listed in the order prescribed by `PEP 8
+# <http://www.python.org/dev/peps/pep-0008/#imports>`_.
+#
+# Standard library
+# ----------------
+import collections  # defaultdict
+import glob  # glob
+import json  # dumps
+import pathlib  # Path
+import sys  # platform
+import urllib.parse  # urlparse
+
+# Third-party imports
+# -------------------
+# We assume a previous call to ``xsltproc`` has already verified that lxml is installed.
+import lxml.etree as ET
+import lxml.ElementInclude
+
+# Local application imports
+# -------------------------
+# None.
+
+
+# Mapping
+# =======
+# Build a mapping between XML IDs and the resulting generated HTML files. The goal: map from source files to the resulting HTML files produced by the pretext build. The data structure is:
+#
+# .. code::
+#   :number-lines:
+#
+#   path_to_xml_id: Dict[
+#       # A path to the source file
+#       str,
+#       # A list of XML IDs in this source file which produce HTML files.
+#       List[str]
+#   ]
+#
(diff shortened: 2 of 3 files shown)


bjones1:
@oscarlevin, @StevenClontz any thoughts/feedback before I convert to a full PR?

In terms of timing (see https://github.com/PreTeXtBook/pretext/pull/1867#issuecomment-1349185699), I can't think of anything. The CodeChat System still works without a mapping file, although having the mapping obviously helps. Comments on that?

oscarlevin:
Looks good to me.  I agree that we don't need to worry about timing here.  A post on dev announcing the removal of the 'html-with-mapping' format of the `pretext/pretext` script might be in order just in case someone is using that.

bjones1:
I'll send a message after this is merged. (The other PR -- https://github.com/PreTeXtBook/pretext/pull/1868 -- is already merged.)

pretext/codechat.py
```diff
@@ -0,0 +1,110 @@
...
+                path = path[1:]
+            # Use ``resolve()`` to standardize capitalization on Windows.
+            stdpath = pathlib.Path(path).resolve()
+            # Make this path relative to the project directory, to avoid writing potentially confidential information (username / local filesystem paths) to the mapping file, which might be published to the web.
+            relpath = stdpath.relative_to(project_path)
+            # Add this XML ID to others for this path.
+            path_to_xml_id[str(relpath)].append(xml_id)
+
+    # Save the result as a JSON file in the ``dest_dir``.
+    (pathlib.Path(dest_dir) / "mapping.json").write_text(json.dumps(path_to_xml_id))
```
StevenClontz, line 110: Should this be hidden (`.mapping.json`)?

pretext/codechat.py
```diff
@@ -0,0 +1,110 @@
...
+#
+#   path_to_xml_id: Dict[
+#       # A path to the source file
+#       str,
+#       # A list of XML IDs in this source file which produce HTML files.
+#       List[str]
+#   ]
+#
+# This allows a single source file to produce multiple HTML files, as well as supporting a one-to-one relationship. The list captures the order of appearance of the XML IDs in the tree -- element 0 is the first XML ID, etc.
+def map_path_to_xml_id(
```
StevenClontz, line 47: I'd appreciate a minimal test added to the suite so I'll know if I break this later by accident. (-:

pretext/build.py
```diff
@@ -40,6 +40,9 @@ def html(
                 None,
                 output.as_posix(),
             )
+            codechat.map_path_to_xml_id(
```
StevenClontz, line 43: Do we want creating the mapping.json to be optional? How expensive is the process?

StevenClontz (changes requested):
Looks good, let me know what you think abou these comments.

pretext/codechat.py
bjones1: Sure, I'll be happy to change the name.

pretext/codechat.py
bjones1: Will do!

pretext/build.py
bjones1: The amount of work seems fairy minimal compared to the processing of a typical build to me: it just re-parses the XML, then looks only at XML xml:id attributes. If I can, I'll try collecting some profile data. What do you think is a good test case -- which book should I build to evaluate performance?

bjones1:
Good feedback -- thanks. A couple of thoughts:

1. When writing tests (always a good idea!), I realized that all the source files in the mapping begin with (of course) `source/`. It makes more sense for paths in the mapping to be relative to the source directory, instead of relative to the project directory. How can I ask pretext-cli for the source directory?
2. There's a lot of cruft (imports never used, imports redefined, f-strings with nothing for format, etc.). Would you consider using a linter? Flake8 is my favorite; here's a [discussion of alternatives](https://inventwithpython.com/blog/2022/11/19/python-linter-comparison-2022-pylint-vs-pyflakes-vs-flake8-vs-autopep8-vs-bandit-vs-prospector-vs-pylama-vs-pyroma-vs-black-vs-mypy-vs-radon-vs-mccabe/).

bjones1:
OK, ready for another view.

StevenClontz:
We're now using Black for formatting. As a rule I want to have similar workflows as Runestone uses (I believe that's how black was chosen) - is flake8 what's used on that end? @bnmnetp

bnmnetp:
Yes

StevenClontz:
Technically there's no such thing as a "source" directory, though I think it's bad practice for now to include any files that aren't inside a (nonproper) subdirectory of the parent of the main PTX file, and every project we'll generate with `pretext new` will use `source/main.ptx` as its core file.

StevenClontz (approved):

StevenClontz:
Squashed, merged. @bjones1 would you make a contribution to help with linting down the road? Document what must be done in README, and update our CI to enforce it if possible so we catch non-linted PRs?

PR #372, opened 2022-12-15
Manifest for project.ptx files

### LINKED ISSUE #368: project.ptx Schema?
Is there a schema file for `project.ptx`? Or is there some way to programmatically generate all valid options?

siefkenj:
Here is my initial attempt at a manifest for a `project.ptx` file. I have no idea what all the options are, and what we should make required.

Also, if someone else knows Relax-ng better than I do, they could probably make the schema much fancier.

When completed, this fixes #368

### DIFF
--- a/.github/workflows/test-on-push.yml
+++ b/.github/workflows/test-on-push.yml
@@ -44,6 +44,9 @@ jobs:
       run: |
         trang ./schema/project-ptx.rnc ./tmp-project-ptx.rng
         cmp ./schema/project-ptx.rng ./tmp-project-ptx.rng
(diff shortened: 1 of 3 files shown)


StevenClontz:
Thanks for this effort. We're changing the schema in CLI 2.x, so if you want to coordinate on how to formalize the new schema and integrate it into the CLI, lets sync up sometime. Unfortunately, that change makes this work moot I think.

oscarlevin:
I'm reopening this to use as a starting place for the 2.0 schema.  Having such a schema will allow pretext-tools to suggest completions when editing the project.ptx manifest.

Issue #376, opened 2022-12-18
Inconsistent working directory use?


oscarlevin:
The recent addition of codechat to the code base revealed a bug that only showed up when `pretext build -g` was used.  A `pretext build` would run the `code_chat()` function just fine, because it received `utils.project_path()` as an argument, and this returned the true project path.  But when `pretext build -g` was run, the `utils.project_path()` returned `None`: it was looking for the `project.ptx` file in the temporary directory.

So something is inconsistent here, and to avoid future gotchas, we should button this up.  Eventually.

For now, the bug is fixed in #375

oscarlevin:
Surely fixed by v2.0

PR #379, opened 2022-12-20
Cache project path for working directory util

StevenClontz:
Attempt to address #376

### DIFF
--- a/pretext/utils.py
+++ b/pretext/utils.py
@@ -24,6 +24,7 @@ from . import core
 # Get access to logger
 log = logging.getLogger("ptxlogger")
 
+CACHED_PROJECT_PATH: Optional[Path] = None
 
 @contextmanager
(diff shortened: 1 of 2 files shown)


StevenClontz:
I think what feels "wrong" about this is that `project_path` should be a property of a `Project` object. So instead of having a `project_path` util, maybe we should have an `initialize_project` util, then pass around the `Project` object as needed. I think right now we have an assumption in the CLI that we're only ever working with one Project at a time, so I'm betting things will look broken for someone using us as a library and tries to create several Projects.

StevenClontz:
But I think addressing that would be a major refactor and would suggest merging this as-is for now.

StevenClontz:
Oh gross I used a `global`? I think the problems here are solved by having path set on the `Project` object, so I'm closing.

PR #385, opened 2022-12-26
add codespace bootstrapping

StevenClontz:


.devcontainer/postCreateCommand.sh
```diff
@@ -0,0 +1,16 @@
...
+curl -sSL https://install.python-poetry.org | python -
+poetry install
+poetry run python scripts/fetch_core.py
+poetry run python scripts/zip_templates.py
+
+echo "Install LaTeX"
+apt update && apt install -y texlive texlive-latex-extra texlive-fonts-extra texlive-xetex texlive-science --no-install-recommends
+
+echo "Install sage"
+apt install -y sagemath --no-install-recommends
```
StevenClontz, line 13: See this bug: https://bugs.launchpad.net/ubuntu/+source/sagemath/+bug/1960459

Because the Codespace we create includes Sage 9.0 in its distribution, we're affected.

StevenClontz: It's unclear to me whether its easier to use a different Linux distribution that includes SageMath 9.2+, or build/install a 9.2+ version of SageMath through another mechanism.

oscarlevin: Yeah, I'm not sure.  Let's start with this even if it is broken and see if we can find another option soon.

PR #386, opened 2022-12-26
New book refactor

StevenClontz:
Implementing discussion at https://groups.google.com/g/pretext-dev/c/m--8VF0vmK4/


StevenClontz:
I'm working on fixing the tests for the new default book

oscarlevin:
What do you think about using this, but changing the target to be something like `book-nested` or `book-structured` or `book-modular`.? Then have the default `pretext new` have target `book` that is just all the support files and a `main.ptx` in `source` that contains the minimal required elements of a book, all in that one file.

StevenClontz:
Based on today's discussion I've consolidated this contribution to a single file. We can complicate things with xincludes later.

StevenClontz:
The default project.ptx cannot refer to `ch_first` as `ch_first` may not exist (and doesn't in this book).

oscarlevin:
I've switched the tests "back" to the `demo` template, since that has more features to test.  I made the `book` template even more bare-bones (removing the tempting "ch1" xml:id)

PR #391, opened 2023-01-03
add -x option to build to mimic generate

oscarlevin:


pretext/cli.py
```diff
@@ -364,6 +373,9 @@ def build(target, clean, generate, project_ptx_override: t.Tuple[str, str]):
     consult the PreTeXt Guide: https://pretextbook.org/documentation.html
     """
 
+    # Add xmlid value to project_ptx_override (a tuple of tuples)
+    if xmlid:
+        project_ptx_override += (("target.targets.xmlid-root", xmlid),)
```
StevenClontz, line 378: Need to update `project.ptx` guidance to use `-x` rather than editing `project.ptx` (messing up Git history).

StevenClontz (approved):

pretext/cli.py
oscarlevin: There might be good reason to have a subset build that builds a particular subset, I suppose.  But right, to just develop a certain section, it is a lot easier to just use the `-x`.  You could use that the `web` target if you don't care about knowls.

I'll fix that comment and merge

PR #394, opened 2023-01-10
fix project_path in case it has spaces

oscarlevin:
This should fix issue reported https://groups.google.com/g/pretext-support/c/A4h7OO7zfX4.  Heads up @bjones1, I added a step in the codechat module that cleans up `project_path` so it matches what `urllib.parse` gives when the path contains spaces.  Feel free to modify this on a new PR if you see a safer way to accomplish this.

### DIFF
--- a/pretext/codechat.py
+++ b/pretext/codechat.py
@@ -17,6 +17,7 @@ import json  # dumps
 import pathlib  # Path
 import sys  # platform
 import urllib.parse  # urlparse
+import urllib.request # pathname2url
 
 # Third-party imports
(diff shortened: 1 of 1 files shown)


bjones1:
@oscarlevin, thanks for this fix! However, I'd recommend a slightly different approach. I'll play and submit some revisions.

bjones1:
I don't have access to directly modify this PR, so I posted my PR in #395.

oscarlevin:
Will soon be replaced by #394

PR #395, opened 2023-01-10
Alternative fix to handling spaces in path names

bjones1:
This fix handles spaces in pathname by URL-decoding the path to produce a pathname that matches the filesystem.

It also included a set of improved tests with:

- A space in the project path
- A space in a .ptx file
- An different XML ID than the filename.

 These tests fail under the old approach, but work with this approach.

### DIFF
--- a/pretext/codechat.py
+++ b/pretext/codechat.py
@@ -17,6 +17,7 @@ import json  # dumps
 import pathlib  # Path
 import sys  # platform
 import urllib.parse  # urlparse
+import urllib.request # pathname2url
 
 # Third-party imports
 # -------------------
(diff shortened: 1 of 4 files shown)


bjones1:
This improves #394 with corrected handling of spaces in filenames, plus adds tests for this case.

oscarlevin:
Excellent.  Thanks for the fix.

oscarlevin:
I might have messed this up by first merging my pull request about the same thing, but in any event, I had to remove the bit where I replaced project_path with its version with encoded spaces.  I just pushed that change, which hopefully will make this work.

PR #396, opened 2023-01-10
Revert "Alternative fix to handling spaces in path names"

oscarlevin:
Reverts PreTeXtBook/pretext-cli#395

Actually, I spoke too soon.  I tried your fix and it doesn't work if there is a space in the path to the file.  For example, if the directory holding the project is `new project` then I get this error:

`critical: 'C:\\Users\\oscar.levin\\Documents\\dev\\pretext-cli\\new project\\source\\main.ptx' does not start with 'C:\\Users\\oscar.levin\\Documents\\dev\\pretext-cli\\new%20project'`

Thoughts?


oscarlevin:
Yeah, I decided I don't want to revert.  I'll try to fix the issue and tests on a new PR.

Issue #384, opened 2022-12-22
Consider reverting `pretext view` behavior on PDF formats


StevenClontz:
![image](https://user-images.githubusercontent.com/1559632/209206682-6c3b9a59-87c1-46a2-8cc4-e799bd851f83.png)

The magic open-in-file-explorer is broken in codespaces. Opening the file in codespaces doesn't preview the PDF.

PR #397, opened 2023-01-11
revert overly-clever view command

StevenClontz:


StevenClontz:
Ah, shoot, I thought I was targeting an earlier commit without the spaces test.

Anyway, this is related to #384. Not only is the existing behavior rather annoying when I just want to pull up the PDF in my browser (you cannot assume the user is on a full PC since PreTeXt is usable on an interface that is *only* a web browser), this behavior completely breaks `custom` format targets like we generate for slideshow. This PR will return `pretext view` to its original behavior of simply running a web server that allows users to see their built files in their browser, regardless of target.

StevenClontz:
See discussion at https://groups.google.com/g/pretext-support/c/qLuRxCnEpJU

oscarlevin:
Yeah, okay.  I think long term we need to think about a better way to manage view, given how we saw folks using it at the PEP (hitting view each time after a build, rather than just going back to the tab and refreshing).  But this is fine for now.

Closes #384

StevenClontz:
For VS Code users, we just want a persistent web service I think, with commands to bring up windows for each output target (or warn to build if it doesn't exist).

PR #399, opened 2023-01-14
reimplement interactive previews

oscarlevin:
Once the second pull request (https://github.com/PreTeXtBook/pretext/pull/1896) has been merged, we will implement the test here, and then should be all ready to go on the CLI.  Note that we now run `playwright install` every time interactives are generated, but this only does anything the first time it is run.


oscarlevin:
Decided to merge this so it can be tested out on the nightly build.  Should be independent of the local server feature still in development, although we can't really run tests until that is available.

PR #405, opened 2023-01-24
remove dummy `pretextbook` setup.py

StevenClontz:
This hack seems to no longer be working to track dependencies of the old `pretextbook` PyPI package anyway, so I propose removing this directory and dummy `setup.py` file.


StevenClontz:
This has somehow completely borked the https://github.com/PreTeXtBook/pretext-cli/network/dependents dependencies. I'll investigate.

PR #406, opened 2023-01-24
update view workflow for cocalc

StevenClontz:
Can be improved upon but would like this in a nightly release so I can try more easily in actual CoCalc soon


StevenClontz:
Looks good in CoCalc

PR #409, opened 2023-01-25
Ignore dummy setup.py file overall by flake rather than using line co…

StevenClontz:
Hopefully this restores https://github.com/PreTeXtBook/pretext-cli/network/dependents

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -15,3 +15,4 @@ exclude =
     pretext/core/
     output
     .venv
+    setup.py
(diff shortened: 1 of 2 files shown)


StevenClontz:
No dice sadly (unless it takes time to propagate): https://github.com/PreTeXtBook/pretext-cli/network/dependents

This one is working: https://github.com/StevenClontz/checkit/blob/main/dashboard/setup.py So maybe I drop the exception and just hope no one is silly enough to try to run it.

PR #411, opened 2023-02-01
add pretext/pretext script runner

StevenClontz:
Usage: `pretext devscript [OPTS/ARGS]` aliases  `./path/to/pretext/pretext [OPTS/ARGS]`.

One deficiency I'm aware of: it uses the default `pretext.cfg` and not `project.ptx`'s executables. However, I think since we're now installing `pretext.cfg` into the user's `~./ptx` folder, maybe we should move executables info there anyway. Thoughts @oscarlevin?

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -14,6 +14,7 @@ import platform
 from pathlib import Path
 import typing as t
 import atexit
+import subprocess
 from .config import xml_overlay
 
 from . import utils, templates, core, VERSION, CORE_COMMIT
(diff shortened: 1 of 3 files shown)


StevenClontz:
Worth considering if `core` is the right subcommand. I guess `pretext pretext -h` is cute usage since it's referred to as the `pretext/pretext` script, but probably too cute. Maybe `pretext dev`?

StevenClontz:
Went with `pretext devscript`.

pretext/cli.py
```diff
@@ -145,6 +146,20 @@ def support():
         log.info("No project.ptx found.")
 
 
+# pretext support
```
oscarlevin, line 149: Do you mean pretext devscript?

Issue #414, opened 2023-02-06
/tmp directories not cleaned up after successful build


bnmnetp:
It would be nice if the temp folders were removed after a successful build.

I cleaned up 38GB of temp folders on my staging machine after it ran out of disk space during the Saturday updates.  The culprit took me a while to find but there were hundreds of tmp folders from pretext builds.

StevenClontz:
I believe this is the only culprit (unless it's happening on the [P](https://github.com/PreTeXtBook/pretext) side).

https://github.com/PreTeXtBook/pretext-cli/blob/main/pretext/project.py#L240

We have a context manager that usually handles cleanup - we can just use it on every build rather than just with custom XSL.

StevenClontz:
Oh, well now I remember why we did that.

https://github.com/PreTeXtBook/pretext-cli/blob/main/pretext/project.py#L344

Can you tell me more about the tempdirectories you cleaned up? Because now I'm unsure we caused them on our side.

rbeezer:
The core routines clean-up after themselves, **unless** verbose (`-vv`) logging is requested.  Is the CLI writing these messages into a file?  Always?

There is a module-level variable which accumulates the temp directories.  They could be released.  I'm guessing there is likely a routine for that?

bnmnetp:
I cleaned them all out on saturday, but when I checked this morning there is one directory like /tmp/tmp0x7680gs for each of the PreTeXt books on runestone academy.

These appear to contain the output of the build (not the source) for each book.

```
ls /tmp/tmpyg6yimow/
 _static                    colophon-2.html           sec_count-conc.html          sec_planar.html
 external                   dedication-1.html         sec_counting-addmult.html    sec_polyfit.html
 generated                  dmoi.html                 sec_counting-binom.html      sec_propositional.html
 knowl                      frontmatter.html          sec_counting-combperm.html   sec_recurrence.html
 acknowledgement-1.html     index-1.html              sec_gt-conc.html             sec_seq-arithgeom.html
```

StevenClontz:
Took a mostly random stab at https://github.com/PreTeXtBook/pretext-cli/pull/415 - we'll see if tests pass (-:

oscarlevin:
Somehow forgot to finish merging this.  It is now part of the latest release, 1.5

PR #415, opened 2023-02-06
Attempt to address #414

StevenClontz:


oscarlevin:
It appears that temp directories are being left behind by `pretext generate` as well.

Is there an advantage to keeping the temp directories when `pretext -v debug` is called?

StevenClontz:
It's unclear to me that more verbose output should be hardwired to not cleaning up temp directories. I'd rather it be a separate switch.

oscarlevin:
What about doing something different if there is an error?  Like copying the temporary directory to a "failed" subfolder of output for inspection.  

Or we can do `core.release_temporary_directories()` at the end of the `try` , which would leave the directory if there is an exception.

I'm adding the `core.release_temporary_directories()` to generate now.

rbeezer:
There is a module-level variable with a list of all the temporary directories created.  So the CLI can handle clean-up in any way you think is appropriate.  If you don't tlike the variable, you could add a method to core to just return the list - that might be best anyway.

oscarlevin:
It looks like we can do all the cleanup at once in a single spot (there is an `exit_command` function already to deal with logging).

However, this has two issues that would need @rbeezer's attention.  First, for some reason, latex-image's temp directories don't get deleted on windows, and this throws an error, unless we set `ignore_errrors=True` on 3811 of pretext/pretext.  Second, we get a bunch of unhelpful info messages spit out after our "success" message.  Thoughts about removing this log message (or putting it in debug, which doesn't get read by anything except our log-file writer)?

rbeezer:
>  First, for some reason, latex-image's temp directories don't get deleted on windows, and this throws an error, unless we set ignore_errrors=True on 3811 of pretext/pretext. 

I took a hard look to see what could be different there.  That routine can end by throwing a `ValueError`, but you would have caught that and seen it?

Maybe the run of LaTeX is still holding some file?  `subprocess.run()` is suppose to be better than the alternatives, so maybe not that.

Might you try deleting individual files and see if there is one in particular?

rbeezer:
> Second, we get a bunch of unhelpful info messages spit out after our "success" message. Thoughts about removing this log message (or putting it in debug, which doesn't get read by anything except our log-file writer)?

I find them helpful.  ;-)

You could move your success message?

If it goes into the debug messages it won't ever say anything, since they are not being deleted?

I think a getter in core for the list of temporary directories is the best route for the CLI to manage this.  Then the release function can move into  `pretext/pretext`.

oscarlevin:
Okay, for the release message, I've moved some things around so it is not a big deal that they get reported.  

The `release_template_directories()` is in pretext/pretext, so I don't know what you mean about the getter comment.

Still not sure about the windows error.  It is not that subprocess.run() needs `shell=True`, which fixed issues I've seen like this previously.  I'll keep looking.  Strangely, if you use `ignore_errors=True`, then all the files in the temporary directory get deleted, but the directory itself does not.

rbeezer:
> The release_template_directories() is in pretext/pretext

I see `release_temporary_directories()` defined in pretext/pretext.py.

> so I don't know what you mean about the getter comment.

Add a function to  `pretext.py` that returns the list in `__temps` (`get_temporary_directories()`?).  Then CLI code can call that function and delete the directories when and how it wants to.

oscarlevin:
Gotcha.  If we need to be more clever than we are now, that is what we will do.

I found the issue with not being able to delete temporary directories: a few times, pretext/pretext changes into the temporary directory (to run latex stuff) and doesn't change back out at the end.  Pull request on core coming right up.

rbeezer:
> I found the issue with not being able to delete temporary directories:

Aah ha!  Beautiful.  Thanks for chasing that down.  I'll follow-up  with https://github.com/PreTeXtBook/pretext/pull/1925 in the morning.

PR #416, opened 2023-02-06
Start to upgrade datafile to dataurl

StevenClontz:
I don't know enough about this feature yet, but I know at least this seems to be necessary...

### DIFF
--- a/pretext/generate.py
+++ b/pretext/generate.py
@@ -372,10 +372,10 @@ def datafiles(ptxfile: Path, pub_file: Path, output: Path, params, xmlid_root):
     source_xml = ET.parse(ptxfile)
     for _ in range(20):
         source_xml.xinclude()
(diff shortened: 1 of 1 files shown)


oscarlevin:
Can you explain what the issue is?  I thought that the new datafile element was what would need to get extracted, and `dataurl` was to support a separate feature that was briefly called datafile.  So I don't think any dataurl needs to get extracted at all.

There does seem to be an issue, I think with core, that is not respecting the `generated` URI from the publication file.  The `generate datafile` feature I added on Friday is generating the file and putting it in the right place, but then `pretext build` complains that it cannot find the resource.  I'll look into that part further.

rbeezer:
`datafile` and `dataurl` are two different animals.  So the common treatment is unlikley to be correct.  And `dataurl` won't require any extra attention from the CLI.

It is possible there is some confusion with: https://github.com/PreTeXtBook/pretext/pull/1878

Issue #417, opened 2023-02-06
Don't download headless browsers until an interactive needs to be screenshot


StevenClontz:

oscarlevin:
Merged fix in #421 .  Part of 1.5 release

PR #421, opened 2023-03-06
build qrcodes on all videos; only install playwrite deps when needed

oscarlevin:
This will build qr codes for all video elements (or at least, not stop them from being built by core).

Also a quick-fix for the playwrite dependencies: only install them when `interactive` elements are in source.


oscarlevin:
Closes #417

PR #426, opened 2023-03-29
support output-filename for latex+pdf builds

StevenClontz:


StevenClontz:
This isn't quite ready. I think we need to redesign the API for our build functions to not support both `output_dir` and `output_filename` - we just accept an `output_filepath` that's the responsibility of the user (or the `Project`) to construct, so we can report back that full filepath on the commandline.

StevenClontz:
Okay I think this is ready for review now @oscarlevin. Probably the rest of the `build.py` module needs to be refactored similarly as well, but perhaps can wait for a later PR.

pretext/build.py
```diff
@@ -13,14 +13,15 @@
...
+    output_dir: Path,
     stringparams,
     custom_xsl: Optional[Path],
     xmlid_root,
     zipped=False,
 ):
-    os.makedirs(output, exist_ok=True)
-    log.info(f"\nNow building HTML into {output}\n")
+    output_dir.mkdir(exist_ok=True, parents=True)
+    os.makedirs(output_dir, exist_ok=True)
```
oscarlevin, line 23: Do we need both these lines?

oscarlevin:
Am I reading the code correctly that with this change:
1. If a user supplies an output-filename in the manifest, then that will be the output filename for the pdf and latex output.
2. If a user does not supply an output-filename, then the output filename will be the target's name (so print-latex.tex, by default).

I think this is not what we want.  Currently, the name of the output is given by the xml:id of the root of the document (I think).  That should be the default unless the user overrides it somehow.

rbeezer:
I once used the root/document `xml:id` as the basename.  I believe that now `pretext\pretext` uses the same basename as the master XML file, but that would be `main` for most projects using the CLI.

Using `@label` might be better long-term.  The pre-processor is copying `@xml:id` into `@label` if it is absent.

StevenClontz:
This got stale and I'm refactoring `project.ptx` anyway now, so I'm closing.

PR #442, opened 2023-05-30
Add `install-deps` playwright command

### LINKED ISSUE #436: Interactive previews not built in codespaces
Likely a ports issue: playwright cannot generate previews for interactive elements when running on codespaces.

oscarlevin:
This should fix the issue of playwright not working on some machines because some system managed deps are missing.  Might prompt the user for a password.

Closes #436

### DIFF
--- a/pretext/utils.py
+++ b/pretext/utils.py
@@ -451,6 +451,8 @@ def playwright_install():
     Run `playwright install` to ensure that its required browsers and tools are available to it.
     """
     try:
+        log.info("Attempting to install/update required playwright packages.")
+        subprocess.run("playwright install-deps", shell=True)
         subprocess.run("playwright install", shell=True)
         log.debug("Installed dependencies to capture interactive previews")
     except Exception as e:


StevenClontz:
Design question: with Codespaces being the direction we point non-expert authors, do we want to manage non-Python dependencies via the CLI at all (besides possibly managing the devcontainer)?

oscarlevin:
I agree it is not clear what we should do here.  It doesn't seem to be an issue on Windows anyway (and this extra line doesn't do anything in that case).  For the codespace setup, we should run this command as part of the docker image, so it will be unnecessary there too.  

However, if even an experienced user tries `pretext generate` to get interactive previews on linux, there is a good chance this will fail and not in a nice way.  Twice I have had to google the error message I got and only then got the hint to run `playwright install-deps`.

Note that we already do `playwright install`, which also seems reasonable and necessary.  The `install-deps` is for the dependencies for the browsers you get from `playwright install`.

StevenClontz (approved):
Since we already do `playwright install`, this seems like the right thing to do for now. I'll open a separate issue to consider whether we should do any dependency management within CLI or factor it all out to devcontainer or something else.

PR #463, opened 2023-06-23
implement deploy for multiple projects; fix few bugs

oscarlevin:
This implements the deployment of multiple targets, assuming that they have a `<deploy-dir>` element, and that a folder called `site` is present in the root of the project.

No tests written yet.

I also made a few changes to the recent type-checking code (specifically, some of the `assert isinstance(...)` lines).  I'm not sure if I did this correctly, but with the upcoming refactoring of `project.py` I think this should be fine for now.

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -673,8 +673,15 @@ def view(
     context_settings=CONTEXT_SETTINGS,
 )
 @click.argument("target_name", metavar="target", required=False)
+@click.option(
+    "-s",
+    "--site",
+    required=False,
+    default="site",
+    help="Relative path to a directory containing html for a landing page.",
+)
 @click.option("-u", "--update_source", is_flag=True, required=False)
-def deploy(target_name: str, update_source: bool) -> None:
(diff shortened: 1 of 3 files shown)


oscarlevin:
@bjones1, apparently I messed up the types stuff, since this isn't passing those tests.  Hopefully you can help me with this on Monday.

oscarlevin:
Got this figured out.  The type `assert isinstance` stuff is bewildering to me, but I'm getting the hang of it.

StevenClontz:
In Typescript, the type checker is smart enough to know if I do something like

```ts
foo:string|number = // stuff
if (foo is string) {
   // foo is type checked as a string
} else if (foo is number) {
   // foo is type checked as a number
}
```

Is MyPy that clever too, or do we need to need to litter asserts everywhere we know that a union type must logically be a particular type?

oscarlevin:
I believe it can do at least some of this.  The issue I had was that a function returned None if a parameter was None, otherwise a list.  Then another function checked whether the function didn't return None, which entails that the parameter wasn't None, but mypy didn't deduce that the parameter wasn't None.

PR #464, opened 2023-06-25
Refactoring to more pythonic Project & Target classes

StevenClontz:


oscarlevin:
Heads up, that I've finished the deploy directory stuff, if you are ready to add "deploy" to the refactor.

StevenClontz:
crossref discussion @ https://github.com/PreTeXtBook/pretext-cli/discussions/458

oscarlevin:
Since I won't be at the next hackathon for this, two comments.  First though, I should say this is looking really good and I appreciate the work you've put into this.

1. What is the advantage of using all the getters/setters?  I think that if other parts of the program need to modify value, then having these is useful.  But I don't think we are doing that ever.  That is, once a target is created, it stays the same (in a particular run of the CLI).  Could we put all of the setter logic into the constructor for the class?  That is, when a project or target is instantiated, we attempt to parse the project.ptx file.  For each property we want to set, we either read it from the manifest (or publication file) or set it to the default in case it is not present.  Before doing this, we create the ElementTree in 2.0 structure from an old manifest, if needed.
2. It looks like the current `build` function takes in a lot of parameters.  Should we just pass it the target (and maybe xmlid from the CLI)?

bjones1:
I agree that I'd prefer simple class member access instead of getters/setters, instead pushing the process of validation and import to Pydantic (see #474).

oscarlevin:
I think the point is that no matter how the target or project is created, all its properties are set at the time it is created.  Thus there is no need for a setter.  But perhaps I am missing something here.


Cheers,
Oscar


Oscar Levin,  Associate Professor

Department of Mathematical Sciences

University of Northern Colorado

Ross Hall 2240D; (970) 351-2380

Schedule Student Hours: https://bit.ly/bookoscar

Zoom: https://unco.zoom.us/my/oscarlevin

***@***.***?anonymous&ep=signature>         Book time to meet with ***@***.***?anonymous&ep=signature>
________________________________

StevenClontz:
Quick mockup of the kind of thing someone might want to do:

```
p = Project()
# right now this returns None but should return the target so
t = p.new_target("web", "html")
for id in list_of_xml_ids:
    t.output = Path("output", "web", id)
    t.build(xmlid=id)
```

StevenClontz:
Or more importantly:

```python
p = Project()
runestone = p.target("runestone")
if runestone is None:
    runestone = p.target("web")
if runestone is None:
    raise Exception
else:
    runestone.publication = "/path/to/runestone/publication.ptx"
runestone.build()
```

BTW @bjones1, writing that out made me decide that `has_target` is more natural:

```python
p = Project()
if p.has_target("runestone"):
    runestone = p.target("runestone")
elif p.has_target("web"):
    runestone = p.target("web")
    runestone.publication = "/path/to/runestone/publication.ptx"
else:
    raise Exception
runestone.build()
```

bjones1:
Code like this:

``` Python
p = Project()
# right now this returns None but should return the target so
t = p.new_target("web", "html")
```

would be if using Pydantic

``` Python
p = Project(target=[Target(name="web", format="html")])
```

Does that fit with your design goals?

StevenClontz:
Thanks for syncing up earlier @bjones1. In 687a0a5 I confirmed my Pydantic classes are validating the examples I created for our tests. But as we discussed, I'm happy for you to refactor things so that our Project and Target themselves subclass `pydantic_xml.BaseXmlModel` directly. We'll sync up at 4pm Central as planned to check in. Thanks!

bjones1:
Fantastic, thanks! I'll work on this.

StevenClontz:
(there seems to be a bug in our CI that causes tests to run twice here, once for push, once for PR?)

StevenClontz:
I assume https://github.com/PreTeXtBook/pretext-cli/blob/main/.github/workflows/test-on-push.yml#L3 should remove the "push" if we want testing on PRs and before PyPI deployments?

oscarlevin:
I'm fine with removing the test on push part, assuming we only ever merge pull requests into main, instead of doing quick updates to main directly (which I am guilty of).

PR #465, opened 2023-06-25
WIP refresh to codespaces

StevenClontz:


StevenClontz:
A weird quirk: the `postCreateCommand` doesn't seem to run. But if I try to `View Creation Logs` it says I need to connect to a codespace, then runs the `postCreateCommand` (and then I can view the creation log).

StevenClontz:
> A weird quirk: the `postCreateCommand` doesn't seem to run. But if I try to `View Creation Logs` it says I need to connect to a codespace, then runs the `postCreateCommand` (and then I can view the creation log).

I think this is a GitHub bug - the same quirk is happening in another unrelated project as well.

StevenClontz:
> > A weird quirk: the `postCreateCommand` doesn't seem to run. But if I try to `View Creation Logs` it says I need to connect to a codespace, then runs the `postCreateCommand` (and then I can view the creation log).
> 
> I think this is a GitHub bug - the same quirk is happening in another unrelated project as well.

Acktually maybe the bug isn't that postCreateCommand isn't running - it's just that the web app doesn't seem to "connect to codespaces" for some reason until I ask to view the creation logs.

StevenClontz:
I tried using `registry.gitlab.com/islandoftex/images/texlive:TL2023-2023-06-25-full` https://gitlab.com/islandoftex/images/texlive/container_registry/573747?orderBy=NAME&sort=desc&search[]= but I think there's a problem with installing poetry: the default user in that docker container is root, and poetry isn't happy with this:

```
Poetry (1.5.1) is installed now. Great!

To get started you need Poetry's bin directory (/root/.local/bin) in your `PATH`
environment variable.

Add `export PATH="/root/.local/bin:$PATH"` to your shell configuration file.

Alternatively, you can call Poetry explicitly with `/root/.local/bin/poetry`.

You can test that everything is set up by executing:

`poetry --version`
```

I'm not sure how to do this permanent path modification as part of the postCreateCommand. But without doing this, the `poetry install` command fails.

StevenClontz:
This version of the Codespace is viable, though could certainly be improved. All tests are passing through the GUI.

image

In https://github.com/PreTeXtBook/pretext-cli/pull/465/commits/07cc7ac3f262be652c22a56e5ef9053c95f629ad I removed pretext-tools and CodeChat, because without a project.ptx in the root they are useless. In any case, the point of this repo is for CLI tools, so it makes more sense to constrain those extensions to a Codespace for their development. (Although if we started writing documentation for the CLI in PreTeXt, it'd make sense to add them back.)

@bjones1 take a looksee for me. A big improvement would be to switch to a container like `oscarlevin/pretext:full` that had less postCreateCommands and a quicker creation time, but note the issues I had with Poetry when I tried it.

StevenClontz:
Since Bryan's out today, switching to @oscarlevin for review.

.gitignore
```diff
@@ -3,9 +3,6 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-# Personal configuration
-.vscode
```
oscarlevin, line 7: I don't think we want to track this.  What if I want a custom keyboard shortcut that you don't like?

.vscode/settings.json
oscarlevin, line 1: These settings can be added to the devcontainer, right?  If we want everyone to have them, I think that is the correct place.

.devcontainer/devcontainer.json
```diff
@@ -1,43 +1,24 @@
...
-          }
-      }
-  
-      // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
-      // "remoteUser": "root"
-  }
\ No newline at end of file
+	"name": "PreTeXt-CLI-dev",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
+	"image": "mcr.microsoft.com/devcontainers/python:0-3.11",
```
oscarlevin, line 6: If we want to test with pretext-full, then we can just use the oscarlevin/pretext:full docker image.  That will simplify the post-create command.

.devcontainer/postCreateCommand.sh
```diff
@@ -2,15 +2,21 @@
...
 curl -sSL https://install.python-poetry.org | python -
-poetry install
+poetry config virtualenvs.in-project true
+poetry install --with dev
 poetry run python scripts/fetch_core.py
 poetry run python scripts/zip_templates.py
+poetry run playwright install-deps
+poetry run playwright install
+
+echo "Update apt"
```
oscarlevin, line 12: For this and the rest of the file: if we switch to the docker image, this is unnecessary (and would be faster)

.devcontainer/devcontainer.json
```diff
@@ -1,43 +1,24 @@
...
-                  "source/main.ptx"
-              ]
-          },
-          "vscode": {
-              "settings": {
-                  "CodeChat.CodeChatServer.Command": "CodeChat_Server"
-              },
-              "extensions": [
-                  "CodeChat.codechat",
-                  "oscarlevin.pretext-tools"
```
oscarlevin, line 36: It isn't crucial to have these, but doesn't hurt (and I get them anyway since I sync my vscode settings).  In particular, pretext-tools also gives snippet and syntax highlighting support, and some limited schema checking.

oscarlevin (commented):
This will be very nice to have.  I'm not sure I will use the codespace, but it should help drive-by contributors.

StevenClontz:
After burning a lot of hours today, this is still not ready.

Part of the problem: https://github.com/PreTeXtBook/pretext-codespace/issues/12 If that was fixed, then I think I can configure this Codespace to work by not using a virtual environment at all, but installing everything globally within the container.

StevenClontz:
My net cast into the waters of StackOverflow: https://stackoverflow.com/questions/76560457

StevenClontz:
Going ahead and merging this so I can safely incorporate in my project refactor branch.

PR #466, opened 2023-06-26
Update CI workflow

### LINKED ISSUE #459: Refactor git workflows?
There's a lot of duplication in the git workflows. Is there some what to refactor these? For example, major, minor, and patch actions are very similar. Is there some way to specify an argument to a github action?

`nightly` seems fairly different than the others.

oscarlevin:
I have consolidated the major/minor/patch deploy actions into one action in which you can select the "level" you deploy at.  

Also fixed a bug in `scripts/prep_nightly.py` that was releasing a nightly even if no new contributions were present. 

This will close #459


StevenClontz (approved):
Looks good to me

PR #485, opened 2023-07-21
Fix: Update the CodeChat Server config file to automatically perform a subset build.

bjones1:


StevenClontz (approved):
Looks fine at a glance (though I'd still prefer a solution where CLI can communicate everything to CodeChat without a config file).

Maybe more short-term: the config file should be gitignored and this default file produced by CLI if it's missing? That way someone can muck with their settings and not affect their git history.

bjones1:
> Looks fine at a glance (though I'd still prefer a solution where CLI can communicate everything to CodeChat without a config file).

I plan providing the `source_path` and `output_path` entries by calling the new project API, but wanted to wait until that's released / a bit more tested. Other pieces may take more thought.

> 
> Maybe more short-term: the config file should be gitignored and this default file produced by CLI if it's missing? That way someone can muck with their settings and not affect their git history.

Until these changes are made, it should stay in git history so edits to the `source_path` and `output_path` are preserved. Creating this file if it doesn't exist certainly makes sense.

PR #486, opened 2023-07-21
Fix: Correct source path of legacy projects.

bjones1:


StevenClontz:
I think output, xsl, etc. also need these changes.

bjones1:
@StevenClontz, this is done and ready for another review.

PR #495, opened 2023-07-26
Fix: Run tests only on a PR; do this on any branch.

bjones1:


oscarlevin (changes requested):
If we left in the `push: branches: [ main ]`, that would also test just on pushes to main, right?  We have an agreement not to do any pushes to main, but there is no way in github to prevent an accidental push to main for admins, as far as I know.

Does this make sense?

bjones1:
OK, ready for another review.

oscarlevin (approved):

PR #500, opened 2023-07-27
output_dir/output_file rename

bjones1:
Fix: Make project.py the new __init__.py.
      Add in new output_dir/output_file structure.

Doing this produced a lot more refactoring...


oscarlevin:
I've switched the base branch to `refactor2`.  Will look this over now.

bjones1:
Oops, thanks!

pretext/project/__init__.py
```diff
@@ -505,7 +510,7 @@ class Project(pxml.BaseXmlModel, tag="project"):
     on the disk, and Paths for where to build output and maintain a site.
     """
 
-    ptx_version: t.Optional[str] = pxml.attr(name="ptx-version")
+    ptx_version: t.Literal["2"] = pxml.attr(name="ptx-version")
```
oscarlevin, line 513: I switched this because I was getting a validation error when trying to run `pretext support` using a legacy `project.ptx` (in fact, even if I put `ptx-version="2"` I got the error). 

However, given that we check for this later, do we need this line at all?

pretext/project/__init__.py
```diff
@@ -549,12 +554,11 @@ def parse(
         xml_bytes = _path.read_bytes()
 
         # Determine the version of this project file.
-
         class ProjectVersionOnly(pxml.BaseXmlModel, tag="project"):
             ptx_version: t.Optional[str] = pxml.attr(name="ptx-version")
 
         p_version_only = ProjectVersionOnly.from_xml(xml_bytes)
-        if p_version_only.ptx_version != "2":
+        if p_version_only.ptx_version is not None:
```
oscarlevin, line 561: This was also a recent change.  Don't we want to assume legacy even if `ptx_version="1"`?

pretext/project/__init__.py
bjones1: That's a bug in my code -- what sort of error did you get? Or how can I reproduce it? When I run `pretext support` on a project just created by `pretext new`, everything is OK.

pretext/project/__init__.py
bjones1: Legacy project should not have `ptx_version` in them at all -- if they do have a `ptx_version`, then we assume it's a new project format.

pretext/project/__init__.py
oscarlevin: You are right.  It must have been because of another thing that hadn't been fixed previously that this was throwing the error.

pretext/project/__init__.py
oscarlevin: Fair.  Given the other issue I was having, I think I started considering cases where someone would try to keep the legacy version by specifying a different attribute.

PR #505, opened 2023-07-28
bubble up warning if a user isn't compatible with upcoming v2

StevenClontz:
I believe this is ready for the field in a 1.7.3 release.

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -80,6 +80,9 @@ def main(ctx: click.Context, targets: bool) -> None:
     if (pp := utils.project_path()) is not None:
         try:
             ProjectRefactor.parse(pp)
+            log.info(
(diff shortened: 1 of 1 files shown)


oscarlevin:
This looks great.  I added a report of success when it works, both to buzz-market the 2.0 release, and so we can encourage everyone to run `pretext` for their project using 1.7.3 before they upgrade to 2.0, to see if they are ready.

StevenClontz:
I think adding stuff to support is duplicative as users will we the reports from main.

Maybe report success in support, and report issues in main?

oscarlevin (approved):

oscarlevin:
My thinking for adding the report to support is that it would make it easier for us to see what the problems are when they copy the results of support to the google group (they might not naturally include the part at the top that says to post on support)

PR #506, opened 2023-07-28
clean up pretext build

oscarlevin:
Some initial work, in case anyone is curious where this is headed.

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -13,8 +13,8 @@ import pydantic_xml as pxml
 from .xml import Executables, LegacyProject, LatexEngine
 from .. import constants
 from .. import core
+from .. import codechat
 from .. import utils
-from .. import build
 from .. import generate
 from .. import types as pt  # PreTeXt types
 from .. import templates
@@ -350,14 +350,15 @@ class Target(pxml.BaseXmlModel, tag="target"):
                     out_file=None,
                     dest_dir=self.output_dir_abspath().as_posix(),
                 )
-                # if project_path is None:
-                #     project_path = utils.project_path(ptxfile)
-                # assert (
-                #     project_path is not None
-                # ), f"Invalid project path to {ptxfile}."
-                # codechat.map_path_to_xml_id(
-                #     ptxfile, project_path, output_dir.as_posix()
-                # )
+                # For codechat:
+                project_path = utils.project_path(self.source_abspath())
+                assert (
+                    project_path is not None
+                ), f"Invalid project path to {self.source_abspath()}."
+                codechat.map_path_to_xml_id(
+                    self.source_abspath(), project_path, self.output_dir_abspath().as_posix()
+                )
+                log.debug(f"Set up codechat map using {project_path} as the project path.")
             elif self.format == "pdf":
                 pass
             elif self.format == "latex":
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -107,7 +107,11 @@ class Target(pxml.BaseXmlModel, tag="target"):
         cls, v: t.Optional[str], values: t.Any
     ) -> t.Optional[str]:
         # The HTML and webwork formats allows only an `output_dir`. All other formats produce a single file; therefore, they allow `output_file` as well.
-        if values["format"] in (Format.HTML, Format.WEBWORK) and v is not None and not values["compression"]:
+        if (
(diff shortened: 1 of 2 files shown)


pretext/project/__init__.py
```diff
@@ -342,24 +339,60 @@ def build(
...
+                    pub_file=self.publication_abspath().as_posix(),
+                    stringparams=self.stringparams,
+                    xmlid_root=xmlid,
+                    file_format="html",
+                    extra_xsl=custom_xsl,
+                    out_file=None,
+                    dest_dir=self.output_dir_abspath().as_posix(),
+                )
+                # For codechat:
+                project_path = utils.project_path(self.source_abspath())
```
oscarlevin, line 354: @bjones1, could you look over the next few lines to make sure this is doing what it needs to?  I wasn't sure what the best way to do the project_path in the refactor (might it just be `Path().resolve()`?)

pretext/project/__init__.py
```diff
@@ -341,25 +338,60 @@ def build(
...
+                    pub_file=self.publication_abspath().as_posix(),
+                    stringparams=self.stringparams,
+                    xmlid_root=xmlid,
+                    file_format=self.compression or "html",
+                    extra_xsl=custom_xsl,
+                    out_file=None,
+                    dest_dir=self.output_dir_abspath().as_posix(),
+                )
+                # For codechat:
+                project_path = utils.project_path(self.source_abspath())
```
bjones1, line 353: We should remove `utils.project_path` and replace it with `Project.abspath()`.

pretext/project/__init__.py
```diff
@@ -341,25 +338,60 @@ def build(
...
+                assert (
+                    project_path is not None
+                ), f"Invalid project path to {self.source_abspath()}."
+                codechat.map_path_to_xml_id(
+                    self.source_abspath(), project_path, self.output_dir_abspath().as_posix()
```
bjones1, line 358: Drop `as_posix()` here -- it's not necessary.

pretext/project/__init__.py
```diff
@@ -341,25 +338,60 @@ def build(
...
-            )
+            if self.format == "html":
+                core.html(
+                    xml=self.source_abspath(),
+                    pub_file=self.publication_abspath().as_posix(),
+                    stringparams=self.stringparams,
+                    xmlid_root=xmlid,
+                    file_format=self.compression or "html",
+                    extra_xsl=custom_xsl,
+                    out_file=None,
```
bjones1, line 349: If we're making a compressed HTML file, shouldn't we allow the `output_file` to determine the name of the resulting .zip archive?

pretext/project/__init__.py
```diff
@@ -341,25 +338,60 @@ def build(
...
-                custom_xsl=custom_xsl,
-                xmlid=xmlid,
-                zipped=self.compression is not None,
-                project_path=self._project.abspath(),
-                latex_engine=self.latex_engine,
-                executables=self._project._executables.dict(),
-                # TODO: what if this isn't defined? Should we have a default in the project file instead?
-                braille_mode=self.braille_mode,
-            )
+            if self.format == "html":
```
bjones1, line 341: I would use `if self.format == Format.HTML:` instead.

pretext/project/__init__.py
oscarlevin: I'm not sure what happens in core if both a `dest_dir` and `out_file` are specified.  I suppose that isn't our problem though.

pretext/project/__init__.py
oscarlevin: Something is strange here.  without the `.as_posix()` I get complaints: Argument of type "Path" cannot be assigned to parametry "des_dir" of type "str" in function "map_path_to_xml_id"

pretext/project/__init__.py
oscarlevin: Also, the `@validator` for `output_filename` would need to change in the case that compression is set.

oscarlevin: This should be fixed now.  Please take a look, @bjones1

oscarlevin:
@StevenClontz: I added in the custom build to the refactor, but didn't do anything with the output filename or directory.  Can you double check that my implementation does what you expect?

oscarlevin:
I'm going to merge this so I can add it to work I'm doing on `generate`.  But @StevenClontz, when you get a chance, please double check the `custom` build logic.

PR #509, opened 2023-07-31
revert to previous pdfCropMargins version

oscarlevin:


bjones1:
Thanks for quickly fixing this!

PR #511, opened 2023-08-01
Project server

bjones1:
Add: Support for server URLs.
Fix: Correct output-dir/output-filename tests, add more.

### DIFF
--- a/pretext/project/xml.py
+++ b/pretext/project/xml.py
@@ -1,16 +1,20 @@
 from pathlib import Path
+import shutil
 import typing as t
 from enum import Enum
 import pydantic_xml as pxml
 
 
+# To prevent circular imports, put this here instead of in `__init__`; however, it's not used in the file.
 class Executables(pxml.BaseXmlModel, tag="executables"):
     latex: str = pxml.attr(default="latex")
     pdflatex: str = pxml.attr(default="pdflatex")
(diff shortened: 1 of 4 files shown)


pretext/project/xml.py
```diff
@@ -1,16 +1,20 @@
...
 
+# To prevent circular imports, put this here instead of in `__init__`; however, it's not used in the file.
 class Executables(pxml.BaseXmlModel, tag="executables"):
     latex: str = pxml.attr(default="latex")
     pdflatex: str = pxml.attr(default="pdflatex")
     xelatex: str = pxml.attr(default="xelatex")
     pdfsvg: str = pxml.attr(default="pdf2svg")
-    asy: str = pxml.attr(default="asy")
-    sage: str = pxml.attr(default="sage")
+    # If not specified, use a local executable if it exists; if it doesn't exist, choose `None`, which allows the generation logic to use the server instead.
```
oscarlevin, line 14: I think this is good.  It makes me wonder if we should really be doing checks ourselves for all executables rather than letting core handle that functionality.

oscarlevin (approved):
Looks great.  Thanks.

pretext/project/xml.py
bjones1: Good point. That's easy to add if we want it.

PR #512, opened 2023-08-01
Main merge

bjones1:
Merge with the latest in the `main` branch.


oscarlevin:
Thanks for moving this along.  I just want to confirm the workflow, so we get a nice history:

We will "Squash and merge" at each step.  From `main` to `refactor2`, and from each pull request into `refactor2`.  Then a final "Squash and merge" from `refactor2` to main will include all the code, but a single commit that combines everything.

PR #515, opened 2023-08-01
Fix: mypy errors.

bjones1:


bjones1:
Note that tests fail, but not (necessarily) because of this PR -- we still have lots of other fixing to do.

Issue #513, opened 2023-08-01
Make deploying to Runestone easy


bjones1:
I'd like to head up this effort with PRs to the `refactor2` branch.

bjones1:
Notes from today's drop-in discussion:

- Create a new format: Runestone. This is really just HTML with a few tweaks.
- Try passing a stringparam of `host-platform = runestone`. This might be an easy alternative to creating a new publisher file with the `<platform host="runestone">` tag unnecessary. Otherwise, take the existing publisher file and perform this edit automatically.
- Modify the templates to add a Runestone target.
- Modify the templates to add the following to the `<docinfo>` tag in `main.ptx`:
    - Add a `<document-id>change-this-for-runestone</document-id>` tag, along with comments explaining what content is allowed for this tag. Comment text: "Note Runestone’s preferred practice for the <document-id> is to have a simple lowercase string with no dashes or other special characters. This value is visible to readers who want to register for your course on Runestone outside of any formal setting."
    - Add the following tag:
     
        `<!-- See https://guide-link for valid values of @shelf; one example (below) is "AP Computer Science". -->`
        `<blurb shelf="AP Computer Science">A simple string (no markup is allowed) to describe your book. Think of something like what you would read on a book jacket. This is used only when publishing using Runestone.</blurb>`
- Check the `<document-id>` to make sure it's valid.
- Disallow `@output-dir` for a Runestone target. Instead, automatically use `published/document-id` instead.

bnmnetp:
I like the direction this is headed...

When you say the "existing publisher file" how do we know what that is?  There could be several publisher files.  -- do we assume that we use the one in the web target?

Another thing to consider is that there are people working on a new shell and style for runestone academy books.  That would restore some of the things that people look for in the nav bar and are sorely missing in the current shells.  Once that is known I guess we could programmatically change the style information to use that for academy to ensure we have a consistent look.

bnmnetp:
Today this is done in the publication file inside the html tag.   Here is what I have in my data structures book:

```
<html>
...
<css style="oscarlevin" colors="blue_red" toc="crc" banner="crc" navbar="crc" shell="crc"/>

...
</html>

```

bjones1:
[...]

PR #516, opened 2023-08-04
Add: The Runestone target.

bjones1:
Add a Runestone format to the project file, making it much easier to build for Runestone.

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -100,13 +100,13 @@ class Target(pxml.BaseXmlModel, tag="target", search_mode="unordered"):
     # We validate compression before output_filename to use its value to check if we can have an output_filename.
     compression: t.Optional[Compression] = pxml.attr()
 
-    # Compression is only supported for HTML.
+    # Compression is only supported for HTML and WeBWorK formats.
     @validator("compression")
     def compression_validator(
         cls, v: t.Optional[Compression], values: t.Any
     ) -> t.Optional[Compression]:
-        if values["format"] != Format.HTML and v is not None:
-            raise ValueError("Only the HTML format supports compression.")
+        if values["format"] not in (Format.HTML, Format.WEBWORK) and v is not None:
+            raise ValueError("Only the HTML and WeBWorK formats support compression.")
         return v
 
     # A path to the output filename for this target, relative to the `output_dir`. The HTML target cannot specify this (since the HTML output is a directory of files, not a single file.)
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -32,6 +32,7 @@ log = logging.getLogger("ptxlogger")
 
 class Format(str, Enum):
     HTML = "html"
+    RUNESTONE = "runestone"
     LATEX = "latex"
     PDF = "pdf"
     EPUB = "epub"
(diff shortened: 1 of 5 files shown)


bjones1:
Note that tests fail, but not (necessarily) because of this PR -- we still have lots of other fixing to do.

pretext/project/__init__.py
```diff
@@ -85,61 +86,49 @@ class Target(pxml.BaseXmlModel, tag="target", search_mode="unordered"):
...
+        # When the format is Runestone, this is overwritten in `post_validate`. Make sure it's not specified.
+        if values["format"] == Format.RUNESTONE and v is not None:
+            raise ValueError("The Runestone format's output-dir must not be specified.")
         return Path(v) if v is not None else Path(values["name"])
 
-    # We validate compression before output_filename to use its value to check if we can have an output_filename
+    # We validate compression before output_filename to use its value to check if we can have an output_filename.
     compression: t.Optional[Compression] = pxml.attr()
 
+    # Compression is only supported for HTML.
```
oscarlevin, line 103: And webwork (i.e., webwork-sets), the build target, not the asset

oscarlevin (changes requested):
Looks great.

pretext/project/__init__.py
bjones1: Is this correct?

``` Python
    # Compression is only supported for HTML and WeBWorK formats.
    @validator("compression")
    def compression_validator(
        cls, v: t.Optional[Compression], values: t.Any
    ) -> t.Optional[Compression]:
        if values["format"] not in (Format.HTML, Format.WEBWORK) and v is not None:
            raise ValueError("Only the HTML and WeBWorK formats support compression.")
        return v
```

StevenClontz:
I missed the discussion yesterday, but something to consider if it hasn't already: if the object is for an author to only worry about a working HTML-format target (either the default "web" name, or a custom "runestone" name), and the only time a Runestone version is built is when Brad's book server builds it, would it make sense for us to concentrate on just providing the hooks necessary to modify a working HTML-format target, and let the implementation happen on the book server side?

E.g. the book server gets the book from Github, grabs (via the Project() api) the "runestone" named target if it exists, and uses "web" if not. Then the format can be coerced to HTML (or error if not HTML), a copy of the publication.ptx file can be made, modified to ensure the runestone method is used for html, and the target's publication path can be rewritten to point to this copy. Finally the output-dir is pointed to wherever it needs to be pointed.

bnmnetp:
@StevenClontz you can see a summary here #513 

My favorite option is described there.  An author works with regular html and then deploys to runestone at that point. we either add a Runestone target or make one up programmatically and then copy the publisher file, or just use the publisher file for html and change the platform programmatically .

I think we want to preserve the "old" option of having an explicit runestone target and publisher file for people like Sean who choose to run their own server.  But don't want or need to mess with the author server infrastructure.

I will need to create some kind of api, that accepts the request to deploy, and validates some kind of token (I still want to grant permission to publish on runestone, not just let anyone deploy anything there). On my side I would then handle the details of cloning to the right place, making sure the document-id is valid and available, adding an entry to the library table etc.

StevenClontz:
One other possibility that doesn't require modifying an author's publication file: what would be the result of having a single Runestone publication file that's used for all books hosted on Runestone? If Runestone is the publisher, why are authors allowed to make publication decisions for that platform? (I expect it's not so simple in practice, but I think this might fit with the underlying theory for having a separate publication file in the first place.)

bnmnetp:
@StevenClontz we did talk about that, but there are many options wrt knowls and other things that authors do set.  Oscar has custom CSS and custom XSL (as do runestone RST authors) so I don't think the "one publication file to rule them all" approach is feasible.

bnmnetp:
Having just said that, I would say that custom xsl feels like an author option not a publisher option.  I want to allow custom xsl especially for author/researchers who want to add new interactive elements that may not be relevant to any text but their own and so do not belong in pretext.

pretext/project/__init__.py
oscarlevin: Looks good!

oscarlevin:
Custom xsl could be used to modify the look/feel of a book, making it very much a publisher option.  But it can also be used to support additional features the author has implemented.  So both authors and publishers would be interested in custom xsl.

oscarlevin (approved):
This looks great.  After our discussion yesterday, I think we need to do more, but let's merge this in so we have everything in once place.

StevenClontz:
Just ignore me if I'm beating a horse y'all already killed. :-) But custom XSL is controlled by project.ptx, so that's not precluded by providing an offiical Runestone publication file. 

CSS was maybe what I had in mind earlier when suggesting practice vs theory: if CSS is an option an author should get to choose, why is it in the publication file? If there's a very small subset of publisher choices an author is allowed to make when hosting on Runestone, we could let them set them via either metadata in project.ptx or via the Runestone book server ("choose your publication file" from a finite list, or create a publication file dynamically from options chosen).

I do think that having Runestone be more opinionated about publication settings is a major change from the status quo that can wait though.

PR #520, opened 2023-08-06
Refactor `view` command

oscarlevin:


oscarlevin:
I have done absolutely nothing to improve the `pretext view` command, but it is now working the same way it was previously.

oscarlevin:
I think this is almost done.  I'd love to get your thoughts @bjones1 and @StevenClontz 

The new `view` command will check for an existing server by looking for a `pretext` process that is a child of a `pretext` process.  Since `view` is the only time we use the multiprocessing module, this should only happen when we have started a server.   If so, we just say, go look at the preview (and give a link).

If this doesn't happen, we start a server, mostly as normal.  

Two issues:
1. If the original view command didn't use the standard port (now 8128), then the second time we call view, it will have the wrong port.
2. The keyboard interupt isn't exiting cleanly.  Not sure what I can do about that.

pretext/cli.py
```diff
@@ -600,71 +554,82 @@ def generate(
...
+        log.info("Starting server.")
+        try:
+            server = project.server_process(
+                output_dir=target.output_dir,
+                access=access,
+                port=port,
+                launch=not no_launch,
+            )
+            server.start()
+        except KeyboardInterrupt:
```
StevenClontz, line 622: Shouldn't this be something like

```python
try:
    while True:
        sleep(1)
except KeyboardInterrupt:
    server.stop()
```

StevenClontz: Maybe instead of `while True`:

```python
while server.is_alive():
````

oscarlevin: Yes!  I think I understand what is happening better now too.  This is fixed.

oscarlevin:
I've added a check to see if the requested port is "correct" on an already running server.  It first checks that `localhost:<port>` is reachable, and then since this should be the root of the project directory, it looks for "project.ptx" as one of the files being displayed.

Open to further improvements, but I'm going to merge this for now.

PR #522, opened 2023-08-06
Refactor deploy

oscarlevin:
The `deploy` command is now set up for the refactor, but see #521 for ideas of how to improve this before we release 2.0


oscarlevin:
There is more deploy work to be done, but we should sync all the current work before making more changes.

PR #526, opened 2023-08-08
Docs for a Runestone target

bjones1:
I added these to get yall's feedback. I like the docs I added to the demo article. However, the additions to the article template seem to brief to me -- many authors won't understand this. How verbose can we be?

### DIFF
--- a/templates/article/source/main.ptx
+++ b/templates/article/source/main.ptx
@@ -1,10 +1,6 @@
 <?xml version='1.0' encoding='utf-8'?>
 <pretext xml:lang="en-US" xmlns:xi="http://www.w3.org/2001/XInclude">
   <docinfo>
-    <document-id>change-this-for-runestone</document-id>
-    <blurb shelf="AP Computer Science">
-        A simple string (no markup is allowed) to describe your book.
-    </blurb>
     <macros>\newcommand{\foo}{b^{ar}}</macros>
     <latex-image-preamble>\usepackage{tikz}</latex-image-preamble>
   </docinfo>
--- a/templates/article/source/main.ptx
+++ b/templates/article/source/main.ptx
@@ -1,6 +1,10 @@
 <?xml version='1.0' encoding='utf-8'?>
 <pretext xml:lang="en-US" xmlns:xi="http://www.w3.org/2001/XInclude">
   <docinfo>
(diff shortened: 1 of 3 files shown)


templates/demo/source/docinfo.ptx
```diff
@@ -4,6 +4,29 @@
 <!-- This is where you can define macros and other book-wide -->
 <!--  settings.                                              -->
 <docinfo xmlns:xi="http://www.w3.org/2001/XInclude">
+  <!--
```
oscarlevin, line 7: This is very nice.  I don't know whether it is true that document-id is only used for Runestone though.  @rbeezer , is `document-id` still needed for the catalog and other things?

oscarlevin (commented):
Overall, I like the direction this is going for "demo"

templates/article/source/main.ptx
```diff
@@ -1,6 +1,10 @@
 <?xml version='1.0' encoding='utf-8'?>
 <pretext xml:lang="en-US" xmlns:xi="http://www.w3.org/2001/XInclude">
   <docinfo>
+    <document-id>change-this-for-runestone</document-id>
```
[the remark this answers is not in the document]
bjones1: Sounds good, I'll revert the changes to `article`.

bjones1:
I shortened these descriptions for the book template, trying to take a bit less space but still explain the core meanings of these tags.

templates/demo/source/docinfo.ptx
rbeezer: > is document-id still needed

It was used in the catalog as a way to "claim" one, and ensure global uniqueness as a result.

It might still be used to avoid clashes between books (such as in situations similar to Runestone).

oscarlevin:
Love it.  Maybe not `change-this-for-runestone` but just `change-me` for the docid?

bjones1:
I updated based on Rob's feedback -- how's this? I used `changeme` after realizing that the Runestone instructions say to avoid using dashes.

oscarlevin:
I like it.  Anything else we should add here?

bjones1:
That's all I can think of. I assume this only makes sense for a book, not for an article/slideshow/

oscarlevin:
Yes, I think this is good now.  Thanks!

PR #527, opened 2023-08-09
Change: Instead of a Runestone format, provide a platform="runestone" attribute

bjones1:
Based on yesterday's discussion, this replaces the `runestone` target with a `platform="web/runestone"` attribute, which can only be applied to a `html` target.


oscarlevin (approved):
Fantastic.  I'm very happy with this solution.

bjones1:
Thanks! The drop-in discussions with you and Steven were really helpful in clarifying how to do this.

PR #528, opened 2023-08-09
Project manifest update

bjones1:
I translated the template `project.ptx` files to the new v2 format. How does this look?

### DIFF
--- a/templates/project.ptx
+++ b/templates/project.ptx
@@ -1,51 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-    This file provides the overall configuration for your PreTeXt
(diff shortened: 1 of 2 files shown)


oscarlevin (commented):
It's great how simple the manifest has become.  

I wonder how we should communicate the defaults and options that are possible.  Maybe a comment block that describes the defaults?  That seems better than having a working target with all the defaults already selected (which is what I did with the publication file, now that I think of it).

Of course, we will also need to document this in the guide.

StevenClontz:
I think a comment with a direct permalink to the guide that helps set these features would be better than a commented sample (which may become stale).

Issue #539, opened 2023-08-11
output-dir vs output-filename


oscarlevin:
The CLI expects to be able to pass both an output-filename and output-dir, in cases of pdf or latex build (at least).  However, core assumes that an output filename includes the path to it (so if both arguments are present, it ignores output_dir).  

This would be easy to fix by always passing the concatenation of both to the output_filename argument, but then the CLI and pretext/pretext will have different behavior.

Thoughts on how we should proceed?  @rbeezer?

PR #540, opened 2023-08-11
Fix: Correct handling of output-filename to prepend the output-dir.

bjones1:


bjones1:
Down to two test failures!

They look like Playwright install problems; see https://github.com/PreTeXtBook/pretext-cli/actions/runs/5837587584/job/15833334841?pr=540#step:9:292.

oscarlevin:
Looks great.  This addresses #539 , which I will leave open in case Rob want's to weigh in

PR #542, opened 2023-08-12
Add asymptote method

oscarlevin:
This adds an optional `asy-method` attribute on either project or target, as either "local" or "server" (the latter being the default)


bjones1:
A bit late, but looks good to me!

PR #550, opened 2023-08-12
avoid mutating stringparams

StevenClontz:


oscarlevin (changes requested):
This is a good idea.  Note that I did something similar with `generate_assets`.  Could you compare the two and make the two approaches consistent?  I think I just passed `stringparam.copy()` to core, so when core changes the dictionary, it doesn't persist onto the build that comes after generating.

StevenClontz:
> This is a good idea. Note that I did something similar with `generate_assets`. Could you compare the two and make the two approaches consistent? I think I just passed `stringparam.copy()` to core, so when core changes the dictionary, it doesn't persist onto the build that comes after generating.

Done. Also opened https://github.com/PreTeXtBook/pretext/pull/2043 to address this on core's end. After that, we can do something like https://stackoverflow.com/questions/9819602/union-of-dict-objects-in-python to avoid mutating stringparams at all on our end.

oscarlevin (approved):
Looks good.  Thanks.

PR #551, opened 2023-08-12
move info to get_target so it doesn't show up with has_target

StevenClontz:


oscarlevin (changes requested):
Good idea.  Can you reformat, and then feel free to complete the merge.

StevenClontz:
will do (proposed the change without a codespace so didn't get formatting automatically)

StevenClontz:
I wonder if there's a bot that could propose formatting fixes on PRs that otherwise pass tests...

StevenClontz:
(Something is up with the codespace that is provisioned - I'm getting type errors from Pylance that don't seem to be errors on PRs, e.g.:

image

PR #553, opened 2023-08-12
typed search_mode argument

StevenClontz:


StevenClontz:
Fixes the typing error caught by my devcontainer environment, noted in https://github.com/PreTeXtBook/pretext-cli/pull/551#issuecomment-1676079723

Not sure why this isn't caught in our test suite.

oscarlevin:
Nice.  I was curious about that too.

PR #556, opened 2023-08-13
stage site before deployment

StevenClontz:


StevenClontz:
FYI @oscarlevin I'm working on this now. This will also allow us to nire conveniently write tests for deployment without actually deploying. Am I correct that tests don't yet exist?

oscarlevin:
Ah, I was just about to clean this up and merge.  Good to know.  I'm in a the middle of a discussion post about this.  Maybe slow down so I can share that?

StevenClontz:
Let's sync up at prose.runestone.academy/dropin if you're able to.

StevenClontz:
In addition to tests passing, a spot-check of https://github.com/TeamBasedInquiryLearning/linear-algebra/commit/170e166a06edfca4e93e7b79be6496ca1dee9b58 seems to be working well also.

oscarlevin:
Nice work.  Thanks for taking care of this.

PR #560, opened 2023-08-14
Avoid two validation errors if `Target.name` isn't defined.

bjones1:


oscarlevin (approved):
This looks great.  I agree with the level of detail here; not sure why I wasn't getting the click.Context thing to work.  

Thank you!

bjones1:
My pleasure! Let me know what else needs fixing!

PR #564, opened 2023-08-22
Fix html/platform in book template

StevenClontz:


StevenClontz:
Documentation at [Subsection 44.4.9: HTML Platforms](https://pretextbook.org/doc/guide/html/publication-file-online.html#online-platform-options) is incorrect (@rbeezer). The documentation at [Section 32.2: Publishing to Runestone Academy](https://pretextbook.org/doc/guide/html/sec-publishing-to-runestone-academy.html#sec-publishing-to-runestone-academy) is correct according to Rob at drop-in right now though, and this PR matches that.

oscarlevin (approved):
Looks good.  Thanks for catching this.

bjones1:
@StevenClontz, thanks for fixing this. I remember being confused by seeing it documented in two ways but assumed that either way would work. I'm glad you cleared this up.

PR #565, opened 2023-08-22
Switch to directory of timestamp.log instead of cli.log

### LINKED ISSUE #285: directory of time-stamped logs
Rather than overwriting `cli.log`, we should have a gitignored `logs` directory that has `TIMESTAMP.log` for each run. It's easy to destroy this directory if someone wants to tidy up, but this way a log isn't lost if another cli command is run.

oscarlevin:
This will close #285, but see question there about how long these should persist.

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -118,7 +118,10 @@ def main(ctx: click.Context, targets: bool) -> None:
                 print(target)
             return
         # create file handler which logs even debug messages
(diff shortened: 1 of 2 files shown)


StevenClontz:
> how long these should persist

It's easy enough to delete the `logs` directory manually, and I don't think we're looking at a huge amount of disk space even with thousands of logs.

PR #567, opened 2023-08-24
Fix stringparams to look for stringparam element

### LINKED ISSUE #566: Stringparams not being passed?
I use the setting `<stringparam key="debug.project.number" value="the-way-it-should-be"/>` at https://github.com/TeamBasedInquiryLearning/linear-algebra/blob/main/project.ptx to get project numbers aligned with their siblings. But my build of the 2023 edition and the in-progress 2024 editions have the default asynchronous numbering: [Activity 1.1.1](https://teambasedinquirylearning.github.io/linear-algebra/2023/LE1.html#activity-1)

I'll do some digging to find out when exactly this was affected.

StevenClontz:
Closes #566

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -102,7 +102,7 @@ class Target(pxml.BaseXmlModel, tag="target", search_mode=SearchMode.UNORDERED):
     braille_mode: BrailleMode = pxml.attr(
         name="braille-mode", default=BrailleMode.EMBOSS
     )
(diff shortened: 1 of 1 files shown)


StevenClontz:
We need to add a test that ensures stringparams are not broken in the future (but I just hacked the code here, I don't have a dev environment at the moment).

StevenClontz:
I don't think this fixes the issue. At least, changing the `stringparam` element to `stringparams` with the latest nightly doesn't fix things, so I don't see why this would...

PR #569, opened 2023-08-25
Launch appropriate URL when viewing

StevenClontz:


StevenClontz:
Note that GitHub Codespaces isn't quite working yet, and I think that's on them: https://github.com/orgs/community/discussions/57165

I may implement this suggestion from there, at least for now.

```
if env.CODESPACES == 'true':
  applicationHost = 'https://' + env.CODESPACE_NAME + '-80.preview.app.github.dev`
else:
  applicationHost = 'http://localhost'
```

StevenClontz:
I think this is ready, and should support Codespaces and other users.

oscarlevin (approved):
This looks great!

PR #572, opened 2023-08-25
Maybe fix legacy stringparams

StevenClontz:


StevenClontz:
no bueno but maybe this is a hint for @oscarlevin to try in a proper dev environment

Issue #575, opened 2023-08-27
Update to pydantic-xml 2


bjones1:
While digging for better ways to validate (see #568), I noticed that pydantic-xml has some big version updates, including moving to pydantic v2. I'd like to upgrade our code to use this -- any objections @oscarlevin or @StevenClontz?

Issue #568, opened 2023-08-25
Validate project.ptx on every build


oscarlevin:
With the refactored project, we are much closer to ensuring that project.ptx must be correct.  But it seems like we can still have additional elements that are not supposed to be there.  For example, stringparm is wrong for v2.  We should check for this and warn or fail with a bad manifest.

oscarlevin:
@bjones1 : Do you know if there is a way for pydantic to do this?  To say something like, any other elements in the xml are not allowed?

I suppose the other option is to define a schema and use lxml to validate against it, and warn if that validation fails.

rbeezer:
> use lxml to validate against it

`lxml` validation is very mysterious.  I could never parse the messages when validation failed.  But I guess you could just say "there was a problem"?

bjones1:
Let me do some digging. I don't see a way in pydantic-xml, but I wonder if pydantic allows this.

bjones1:
The short answer: Pydantic-xml doesn't support this in a user-friendly way. Per the [docs](https://pydantic-xml.readthedocs.io/en/latest/pages/data-binding/elements.html#elements-search-mode), there are three ways Pydantic-xml matches elements to the model:
- Strict. This means that elements must be in **exactly** the same order as they are in the model; swapping the ordering produces a validation error. But, adding anything to the XML files also produces an error. We want this, but (IMHO) don't want to require a specific ordering for validation to pass.
- Unordered. This is what we use. However, it ignores elements that it doesn't recognize.
- Ordered. Useless for us.

I updated the code to Pydantic v2 and added a setting to forbid extra model attributes in #576. This helps for API usage, but not for reading in an XML file.

bjones1:
I added a feature request for this: https://github.com/dapper91/pydantic-xml/issues/105.

oscarlevin:
This has since been addressed, via the pydantic-xml update, and #575 among others.

PR #576, opened 2023-08-27
Pydantic v2

bjones1:
Update to Pydantic v2, plus limit extra model attributes (but only for API access, not in reading the project file).

See #575, #568.


bjones1:
My pleasure! At a minimum, keeping up with Pydantic is good. I'm hoping the author of pydantic-xml will consider some of my feature requests (line number errors, validation for extra elements). If so, he'll add it to his v2 library, so this code is important.

PR #585, opened 2023-09-09
Add: Upgrade to pydantic 2.2.0, which reports validation errors on extra attributes and elements.

bjones1:


bjones1:
I had to remove two elements in the legacy test projects: `<pdfcrop>` and `<pageres>`. Are these actually invalid, or should they be allowed?

oscarlevin:
Yes, those executables are stale and should be removed.  

Will this make it impossible to run pretext for anyone with a legacy project that has those extra elements, or some other extra elements?  It would be nice to enforce this for v2.0 manifests, but not for legacy (if that is possible).  

I'm seeing a change to the .lock file, but not to anything else, other than tests, btw.

bjones1:
Yes, it's just a change to the .lock file -- everything else was already in place. Your PR does the right thing to allow legacy projects to avoid this check.

oscarlevin:
Awesome.  Glad you discovered this.  Hopefully everyone who has upgraded to 2.0 since the pydantic-xml fix already had 2.1 installed so didn't get the 2.2 release and hasn't been caught.  I'll push a 2.0.1 bugfix shortly.

PR #587, opened 2023-09-10
Remove pydantic forbid extra elements in legacy

oscarlevin:
The recent upgrade to pydantic-xml (to 2.2.0) adds the feature to forbid extra elements not specified in the model.  This is a great feature for v2, but we have been promising users who had previously had extra elements that their projects would still work.

So I have removed this requirement.  Users who have extra non-read elements should upgrade to v2 manifests soon anyway, and if a legacy manifest is wrong and it is causing issues because of a misspelled element name, then we can deal with that help request just as easily as dealing with the help request when their project no longer compiles.

### DIFF
--- a/pretext/project/xml.py
+++ b/pretext/project/xml.py
@@ -45,13 +45,13 @@ class LatexEngine(str, Enum):
 
 
 class LegacyStringParam(pxml.BaseXmlModel):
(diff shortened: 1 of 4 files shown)


bjones1:
Makes sense to me!

PR #589, opened 2023-09-10
Avoid generating unneeded assets

### LINKED ISSUE #435: Ensure interactive previews are only built when needed
Right now, interactive previews seem to be built no matter the target.  But likely these are not needed for `html`?

oscarlevin:
This will close #435.  It maintains a list of which assets could be used for each format, and filters by those early in the generate_assets method.

Also cleans up constants.BUILD_FORMATS which was left over from legacy and was only used in one help utility.

### DIFF
--- a/pretext/constants.py
+++ b/pretext/constants.py
@@ -19,7 +19,7 @@ ASSETS_BY_FORMAT = {
         "youtube",
         "codelense",
         "datafile",
-        "interactive"
+        "interactive",
     ],
     "latex": [
         "webwork",
@@ -28,7 +28,7 @@ ASSETS_BY_FORMAT = {
         "youtube",
         "codelense",
         "datafile",
-        "interactive"
+        "interactive",
     ],
     "epub": [
         "webwork",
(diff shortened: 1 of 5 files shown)


pretext/constants.py
```diff
@@ -2,17 +2,80 @@
...
-    "kindle",  # TODO: mode of epub rather than separate format?
-    "braille",
-    "html-zip",  # TODO: deprecate
-    "webwork-sets",  # TODO: just "webwork"
-    "webwork-sets-zipped",  # TODO: deprecate
-]
+FORMATS = ["html", "pdf", "latex", "epub", "kindle", "braille", "webwork", "custom"]
+
+# Give list of assets that each build format requires.
+ASSETS_BY_FORMAT = {
```
oscarlevin, line 8: @rbeezer, when you get a chance, can you check that these are indeed the assets that someone might need to generate for each build format?

rbeezer: Those look right to me.  I assume some things are subsumed into others, such as `pg-macros` being part of `webwork`.  `qrcode` part of `interactive`, etc?

Running `pretext/pretext -h` will give you a fairly concise rundown of what core Python can do.

oscarlevin: Awesome.  Thanks!

PR #590, opened 2023-09-13
Fix: Update to pydantic-xml 2.2.1, which now checks wrapped elements.

bjones1:


oscarlevin:
Nice.  Thanks for staying on top of this.

PR #592, opened 2023-09-16

StevenClontz:


oscarlevin:
Tests are failing because of a PR I made on core that messed up the schema.  Once @rbeezer accepts https://github.com/PreTeXtBook/pretext/pull/2071, we can update to that commit.  

Until then, I will revert to a commit prior to the mess that should resolve at least that issue.  Give me a couple minutes.

PR #593, opened 2023-09-16

StevenClontz:
rebased upon main, replaces #592

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -121,7 +121,6 @@ def main(ctx: click.Context, targets: bool) -> None:
         project.generate_boilerplate()
         if targets:
             for target in project.target_names():
-            for target in project.target_names():
                 print(target)
             return
         # create file handler which logs even debug messages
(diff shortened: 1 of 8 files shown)


rbeezer:
PreTeXt already does some file management (aka directory management) and I would prefer the two would not be confused.  Could you just as well use something like:
```
# Managed by the PreTeXt-CLI
```
or similar to make it clear this is a front-end function?

StevenClontz:
I'm failing at Git management today. Closing and making a clean commit based upon main

StevenClontz:
> PreTeXt already does some file management (aka directory management) and I would prefer the two would not be confused. Could you just as well use something like:
> 
> ```
> # Managed by the PreTeXt-CLI
> ```
> 
> or similar to make it clear this is a front-end function?

In #594 I've amended this to look for `<!-- Managed automatically by PreTeXt authoring tools -->`

rbeezer:
`<!-- Managed automatically by PreTeXt authoring tools -->`

I like it!  Thanks.

Issue #582, opened 2023-09-07
Refresh .gitignored config files on every run of the CLI marked as managed by CLI


StevenClontz:
Rather than require users to maintain `codechat_config.yaml`, we can just generate it for them every time the CLI is called.

We can add a safeguard that only overwrites the YAML if it contains a comment saying it's auto-generated. That way folks can configure it manually as well.

StevenClontz:
Rather than requiring a manual `pretext init -r`, or tracking hashes of files we know we've generated, maybe this is a simpler solution. We can have a line like

```xml
<!-- Managed by PreTeXt-CLI -->
```

or

```python
# Managed by PreTeXt-CLI
```

or

```json
"managed_by_pretext_cli": true,
```

that we include in every file we want to keep updated for our users. Then, they can edit these manually if they want, but if so, they are responsible for keeping them updated. Otherwise, they Just Work and whenever we want to change, say, the .devcontainer, we just overwrite it the first time `pretext` is run after a `pip install --upgrade pretext`.

bjones1:
I like it!

oscarlevin:
This is completed in #594

PR #594, opened 2023-09-16

StevenClontz:
Replaces #593


pretext/project/__init__.py
```diff
@@ -906,7 +908,7 @@ class Project(pxml.BaseXmlModel, tag="project", search_mode=SearchMode.UNORDERED
     """
 
     model_config = ConfigDict(extra="forbid")
-    ptx_version: t.Literal["2"] = pxml.attr(name="ptx-version")
+    ptx_version: t.Literal["2"] = pxml.attr(name="ptx-version", default="2")
```
oscarlevin, line 911: Won't this make any manifest without a `ptx-version` attribute be recognized as version 2?

oscarlevin:
This all makes lots of sense for the files other than `project.ptx`, which I think people will modify fairly regularly.  A few ideas:

1. Instead of creating a time-stamped version of the *new* file, why not copy the old file to `bk.old-file`?
2. I don't love the idea of a magic comment being the thing we are checking for.  Instead, what if on every stable release, we hash the template versions of these files and add them to a static pickled dictionary that we ship.  Then we only completely replace those files that match any one of the hash's (indicating that the user has not modified the file shipped, or has happened to modify it exactly into a different static version).

StevenClontz:
> This all makes lots of sense for the files other than `project.ptx`, which I think people will modify fairly regularly.

Disagree somewhat - I think most authors start writing PreTeXt to get web+print with a single source. I think `project.ptx` will be the first file to change, but there's still reason to manage it for casual users until they delete the magic comment.

> Instead of creating a time-stamped version of the new file, why not copy the old file to bk.old-file?

Strong disagree: this is happening everytime `pretext [anything]` is run.

> Then we only completely replace those files that match any one of the hash's

This is what we discussed a while back. It makes sense, but is maybe more work (we have to remember to get these hashes every time we tweak things, which seems brittle). In https://github.com/PreTeXtBook/pretext-cli/issues/582#issuecomment-1718023904 I suggested this alternative which @bjones1 seemed to like.

pretext/project/__init__.py
StevenClontz: I think we have different logic when running `Project.parse()`. The goal here is to let `Project()` work without raising an error.

oscarlevin:
>> Instead of creating a time-stamped version of the new file, why not copy the old file to bk.old-file?

> Strong disagree: this is happening everytime pretext [anything] is run.

Wait, the proposal is to stealth update configuration files every time a user runs pretext in any way?  So if a user has modified a file or removed the magic comment, then they will be getting as many timestamped backups as times they run pretext?  Or would the timestamped backups only occur when `pretext init --refresh` is called?  That's better (and that would be the only time I would do the `bk.oldfile`), but I don't like the idea of doing things the user doesn't ask for.  Further, the only time any files would need to be updated is after an upgrade.

In terms of the hash vs magic comment: I think that the brittleness of every user needing to learn about the magic comments outweighs the brittleness of our automatic workflow "remembering" to rehash on every release.

Thinking more broadly, what if we implemented a `pretext upgrade` command that checked pypi for the most recent version, pip upgrades if there is a new version, and updates configuration files.  We can check for a new version on every run (or once every 7 days) and prompt the user to encourage them to upgrade.

StevenClontz:
For clarity, the workflow I'm trying to implement:

- "Boilerplate" files:
    - requirements.txt
    - project.ptx
    - codechat_config.yaml
    - .gitignore
    - .devcontainer
- User runs `pretext [anything]`, 
    - Before basically anything else, all boilerplate files containing `<!-- Managed automatically by PreTeXt authoring tools -->` (including non-XML files, which have e.g. `# <!-- Managed automatically by PreTeXt authoring tools -->`) are overwritten with the current version of our templates
    - Boilerplates missing this comment have nothing done to them.
- User runs `pretext init --refresh`
    - Boilerplates missing this comment get a `[filename].[timestamp].[ext]` that the user can consider as a replacement.

StevenClontz:
> what if we implemented a `pretext upgrade` command

I'm in favor of this, but it seems harder and I don't want to mess with it here.

StevenClontz:
> I think that the brittleness of every user needing to learn about the magic comments

I don't think this will be the case.

For new users, we create those comments for them, and they Just Work(TM) until they have a use case to edit the file, in which case these comments are up top and have instructions to delete them in order to get manual changes.

For existing users, this is a one-time change: decide if you care about your `.devcontainer.json`. If you don't, delete the file and run `pretext init --refresh` so you never have to care about it again.

StevenClontz:
@oscarlevin I plan to lurk in PROSE office hours today (during my student office hours, but no one is here right now) if you want to discuss

StevenClontz:
Action items:

- [ ] Use `*.bak` files rather than timestamps of new.

pretext/project/__init__.py
```diff
@@ -1242,3 +1243,56 @@ def deploy(
...
+        update_requirements: bool = False,
+        resources: t.Optional[t.List[str]] = None,
+    ) -> None:
+        """
+        Generates boilerplate files needed/suggested for
+        a PreTeXt project. Existing files will be overwritten
+        as long as they have a comment confirmed they are managed
+        by this library and not the user. If `skip_unmanaged` is
+        `False`, then files without this comment will be generated
+        with a timestamp in the filename.
```
oscarlevin, line 1260: Can you fix this docstring to remove the timestamp bit?

oscarlevin (changes requested):

oscarlevin:
Do you want to add the magic comment to the `project.ptx` in the slideshow template?

StevenClontz:
> Do you want to add the magic comment to the `project.ptx` in the slideshow template?

Oh yeah, as you realized, this *shouldn't* be done.

oscarlevin:
Turns out we don't want to manage the project manifest for slides, because we need those customizations.

StevenClontz:
Cool! Even if this isn't in a new release, I can use a nightly to record some tutorials soon.

PR #595, opened 2023-09-18
Fix: Use a model validator to set project's _path, _executables.

bjones1:


oscarlevin:
I don't really understand this, but I trust you know what you are doing @bjones1.  Merging in now.

bjones1:
Thanks! This is some improvements based on issues I posted to the author of pydantic-xml, which helps write better, more maintainable code. It doesn't change any of the behavior, though.

PR #609, opened 2023-10-06
Fix `pretext build --no-generate`

StevenClontz:


StevenClontz:
This is ready. Feel free to push back on the removal of `pretext build -g`. Not that it was working as intended, but it's technically a breaking change and might necessitate a `2.1` release at least. Another option would be to leave `-g` as a deprecated option and log a warning that it's non-functional so at least `pretext build -g` doesn't raise an error.

pretext/cli.py
```diff
@@ -419,12 +409,6 @@ def build(
         log.debug(e, exc_info=True)
         return
 
-    # Call generate if flag is set
-    if generate:
```
StevenClontz, line 423: This was running even when `no_generate == True`. But we're not really respecting what asset was requested on the command line anyway. I think if you want to manually generate assets, just run `pretext generate`.

rbeezer:
I'm generally a fan of gracefully deprecating things that change for some period of time.

StevenClontz:
Rob is of course right that we should leave `-g` for now, so I've added it back. It does correctly respect the `no_generate` flag, though I only implemented its actual behavior ([by setting `flag_value` the `is_flag=False` was being ignored](https://click.palletsprojects.com/en/8.1.x/options/#feature-switches)) rather than the intended behavior.

oscarlevin:
I would rather keep the `-g` flag, but very happy you fixed my oversight of not converting it to a flag (which I think we had planned on, I just forgot to do).

`pretext generate` has the option to only generate changed assets.  `pretext build` will have the option to generate no assets, only changed, or all assets.  This seems like reasonable flexibility to offer.

Issue #610, opened 2023-10-06
factor out type checking from rest of tests


StevenClontz:
Like our formatting check, I don't think we need to run mypy along with our other tests. It should be a separate workflow. In particular, it'd be nice to know if tests are passing even if types have issues.

oscarlevin:
Should be fixed in #612

PR #612, opened 2023-10-06
Updates and refactor action

oscarlevin:
Add pymupdf dependency, update core, and refactor test-on-pr to move mypy check to format job


oscarlevin:
Will close #610

PR #614, opened 2023-10-08
don't regenerate identical boilerplate, closes #613

### LINKED ISSUE #613: don't rewrite `.devcontainer.json` unless there's something to change
By overwriting `.devcontainer.json` VS Code wll prompt Codespaces users to rebuild their container, even if the file is identical to the old file. So we should not touch the file unless it actually needs to be overwritten.

StevenClontz:
closes #613


oscarlevin:
Good idea!

PR #620, opened 2023-10-12
Manage asset directories for latex and custom builds

### LINKED ISSUE #619: `latex` format fails to copy external directory
To replicate:

```bash
pretext new
cd new-pretext-project
mkdir assets
touch assets/foo.txt
pretext build print-latex # no external directory is copied!
```

oscarlevin:
Will close #619


StevenClontz (approved):

StevenClontz:
I mentioned in another thread that we should use the `core.manage_directories` but you've added a feature that doesn't make sense there: ignoring pickles. So it makes sense to do this here.

That said, maybe this reveals the need to put our pickles somewhere outside the folder moved as part of `generated`?

PR #649, opened 2023-12-01
Use/demo CRC in template publication.ptx

StevenClontz:
We incorrectly say some of the attributes are not implemented, now that CRC is a thing. I think CRC is nicer so I suggest enabling it by default for new projects.

### DIFF
--- a/templates/book/publication/publication.ptx
+++ b/templates/book/publication/publication.ptx
@@ -106,10 +106,10 @@
     <css
         style="default"
         colors="blue_red"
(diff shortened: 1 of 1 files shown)


oscarlevin (changes requested):
I think we should include the attributes that are now available, but should set defaults by...default.  That has been the philosophy of the publication file template so far, and I don't feel comfortable being opinionated here.

oscarlevin (approved):
Looks good.  Thanks.

PR #674, opened 2024-01-28
improve static file replacement; use nojekyll

oscarlevin:
This will fix the issue with github pages not seeing the static files (previously jekyll assumed that anything starting with an underscore should be ignored)


StevenClontz:
Oh no! This explains the weirdness - I've been sticking `.nojekyll`s in my `site` directory just to speed up deploy times. Thanks for the catch!

PR #704, opened 2024-03-06
Respect project's requirements.txt

### LINKED ISSUE #676: Codespace should respect `requirements.txt`
I'm getting a lot of feedback from folks who rebuilt their Codespace recently about things suddenly not working. The issue stems from the fact that the Codespace installs the latest PreTeXt, and not the PreTeXt they specify in `requirements.txt`. So something that works using version `2.3.4` but is broken in `2.3.5` will suddenly stop working for the user when they rebuild their codespace. I think this is undesired behavior.

I suggest we run `pip install -r requirements.txt` on the creation of a codespace.

StevenClontz:
Closes #676

### DIFF
--- a/.github/workflows/codespace-template.yml
+++ b/.github/workflows/codespace-template.yml
@@ -33,7 +33,7 @@ jobs:
       - name: pretext new book
         shell: bash
         run: |
-          rm -f project.ptx
+          rm -rf *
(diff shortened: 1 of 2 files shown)


oscarlevin (commented):
Nice fix.

PR #706, opened 2024-03-19
Prefix tmp directories with pretext_

siefkenj:
This PR prefixes temporary directories that pretext creates with `pretext_` to make them easier to manage during debugging.

However, this PR seems to miss the creation of some directory, since my `/tmp` is littered with `tmp****` files that persist after `pretext build` has finished and contain the contents of the build.


StevenClontz:
The `pretextbook/pretext` script creates a lot of temp directories as well - it'd be good to add a PR there as well.

StevenClontz (approved):

rbeezer:
Those directories only persist when the "-vv" flag is used.

Or, there is an easy way to remove the accumulated (global) list of every one that is created.

PR #713, opened 2024-04-05
add github actions to boilerplate

StevenClontz:


.github/workflows/test-on-pr.yml
```diff
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  format:
+  format-and-types:
```
StevenClontz, line 11: because types keep failing here, not just formatting

pretext/cli.py
```diff
@@ -324,7 +324,7 @@ def new(template: str, directory: Path, url_template: str) -> None:
     "files",
     help="Specify file to refresh.",
     multiple=True,
-    type=click.Choice(constants.PROJECT_RESOURCES, case_sensitive=False),
+    type=click.Choice([r for r in constants.PROJECT_RESOURCES], case_sensitive=False),
```
StevenClontz, line 327: changed `constants.PROJECT_RESOURCES` to an object where keys are filenames and values are paths

pretext/constants.py
```diff
@@ -143,10 +144,12 @@
...
 }
 
-PROJECT_RESOURCES = [
-    "project.ptx",
-    "codechat_config.yaml",
-    ".gitignore",
-    ".devcontainer.json",
-    "requirements.txt",
-]
+PROJECT_RESOURCES = {
```
StevenClontz, line 147: see?!

templates/test-build.yml
```diff
@@ -0,0 +1,35 @@
+# <!-- Managed automatically by PreTeXt authoring tools -->
+# (delete the above line to manage this file manually)
+
+name: Test Build
+on:
+    # Runs on pull requests, when a pull request is merged into main,
+    # and when a commit is otherwise pushed to main.
+    push:
```
oscarlevin, line 8: I would rather have this one not automatically run by default either.  I'm worried that a new user will be experimenting, get emails saying an action failed, and get worried.

oscarlevin (changes requested):
I added one suggested change.  Besides that I think this looks really good.

Eventually we should write a better boilerplate for a README that says what all the extra files are that the CLI creates.  But that's for another time.

templates/test-build.yml
StevenClontz: Can we try it like this? If you're right, we can push an update that automatically replaces this with your suggested behavior.

StevenClontz: And it'll help a lot with support. "Why won't my book build?" With this we just ask for the repo and check out the action and artifact.

oscarlevin (approved):

PR #729, opened 2024-05-21
`warning: 'NoneType' object has no attribute 'split'`

StevenClontz:
Caught at https://github.com/TeamBasedInquiryLearning/precalculus/issues/230

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -210,7 +210,7 @@ def support() -> None:
         project = Project.parse()
         project.init_core()
 
-        for exec_name in project.get_executables().dict():
+        for exec_name in project.get_executables().model_dump():
             if utils.check_executable(exec_name) is None:
                 log.warning(
                     f"Unable to locate the command for <{exec_name}> on your system."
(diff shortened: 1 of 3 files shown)


StevenClontz:
Okay, there are two issues here.

1. We have the ability to save the `sage` executable as `None`, which is supposed to rely on some sort of server solution. However, we're failing to do this, and we're running `split()` on the `None` object.
2. Our codespace no longer ships with Sage.

StevenClontz:
This is an upstream bug in core PreTeXt - `core.get_executable_cmd(exec_name)` raises this error if `__executables[exec_name]` returns `None`.

StevenClontz:
Well, define "bug". We're the one that is setting that value to be `None` at https://github.com/PreTeXtBook/pretext-cli/blob/4ce468b30a74fe378bfd911c6d64b44f6819c82f/pretext/project/__init__.py#L1272 My guess is that this is not Rob's intended behavior.

StevenClontz:
The NoneType error has been fixed as of 2a51a75.

But the bigger issue is that we're no longer shipping SageMath in our Codespace.

oscarlevin (approved):
Looks good to me.

Issue #336, opened 2022-08-08
click-log and stderr


oscarlevin:
As mentioned in #334 , I have proposed switching to click-log instead of click-logging, to address #331 .  However, this introduces another strange thing, namely that all logs are sent to stderr instead of stdout for info and debug (and I think stderr for warnings, errors, and criticals).  If we are okay with this, then we should go ahead, but if not, then I think it means abandoning both click-log and click-logging.  I like the idea of getting rid of a poorly maintained dependency, but it would require a little work.

This would require creating our own formatter (I like the colors and lowercase levelnames, but I think I could replicate what they have easily).  More complicated would be to implement the simple verbosity option that we have now (the `-v debug` behavior).  However, this is an opportunity to maybe do something better.  We could follow the core pretext method of `-v` for info messages, `-vv` for debug and nothing for only warnings and above.  Or keep nothing for info and above, `-q` for errors and above, and `-v` for debug.  We can probably also set this up so putting those flags anywhere works, not just after `pretext`.  

Since this is a feature breaking change, I thought I would through it out before we ship 1.0.  What do you think @StevenClontz?

StevenClontz:
`-v` and `-vv` are non-semantic commands, I much prefer `-v DEBUG|INFO|...` and especially like the default set to info as we do currently.

I'm not excited about rolling our own logging.

Looking at https://github.com/click-contrib/click-log/blob/master/click_log/core.py#L33 I think we could roll our own ClickHandler and override the logger's handlers after a basic_config.

StevenClontz:
Here's some context for the stderr btw: https://github.com/click-contrib/click-log/issues/15

oscarlevin:
Yeah, I saw that issue, and that last comment is what makes me suspicious.  I will spend another day thinking about how to override the basic_config.  I know how to make our own handler, but that would give double messages I think.  click-log overrides the default emit() function of logging to use click.echo and passes the stderr thing.  In any event, I think this is fine for now (shouldn't delay 1.0), since we are not changing the interface.

PR #731, opened 2024-05-27
logs below error sent to stdout

### LINKED ISSUE #660: fix stdout vs stderr
CodeChat color-codes based on standard output vs errors:

image

At some poine we should find a way for CLI to only use stderr when we have warnings and errors.

oscarlevin:
This changes the logging of debug, info, and warning to go to stdout.  Actually, everything now goes to stdout, but at the end of the build, when we re-report errors and critical messages, those will go to stderr.

If we wanted warning to also go to stderr, we could either report those warning at the end, or would need to add some more logic to separate which messages go where.

This will close #660.  See also #336

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -46,7 +46,7 @@ mh = logging.handlers.MemoryHandler(
     capacity=1024 * 100, flushLevel=100, target=sh, flushOnClose=False
 )
 mh.setLevel(logging.ERROR)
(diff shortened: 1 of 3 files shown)


oscarlevin:
@StevenClontz , when you get a moment, can you look this over and see if you agree with this approach?

StevenClontz:
hopefully will tackle this tomorrow during drop-in

StevenClontz (approved):

Issue #733, opened 2024-05-31
Subtle issues with capitalization of filenames


StevenClontz:
We found an issue where when a filename was changed from having an uppercase to a lowercase letter, it did not update the file upon `pretext deploy` in the stage.

StevenClontz:
As a quick remedy, we will simply delete the `stage` directory for every `stage_deployment`, which should prevent the issue of failing to update files when capitalization changes.

PR #734, opened 2024-06-01
Clean the stage directory on each deployment

oscarlevin:
To ensure that there is not extra files in the stage directory, and to avoid issues with file name capitalization, every time deploy is called, we delete the stage directory and copy everything fresh.


oscarlevin (commented):
Will close #733

PR #740, opened 2024-06-15
add support to not use knowls (e.g. for subtree previews)

### LINKED ISSUE #738: 404 on xref to an individual exercise
Attempting to create an `xref` to a specific exercise properly generates a link within the book text, however clicking the link causes a 404.

A minimal example can be found here: https://github.com/braughtg/ex-link-pretext-bug

## To reproduce:
- Open https://github.com/braughtg/ex-link-pretext-bug in a Codespace
- Open `source/sect-section-name.ptx` in a CodeChat Preview
- Click on the link to "Checkpoint 1.1.1" at the bottom of the document.

## Observed Behavior:
When clicking the "Checkpoint 1.1.1" link I see the following error message:
image

## Expected Behavior:
The document should jump to the appropriate exercise.

## Notes:
- An `xref` to an `<exercises>` block works fine.
- Pointing at the "Checkpiont 1.1.1" link shows what seems to be a correct URL to an anchor in the current page. However, clicking the link tries to navigate to a `knowl/xref/ex-foss-community-principles.html` file that does not exist.

StevenClontz:
Closes #738

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -408,7 +408,7 @@ def init(refresh: bool, files: List[str]) -> None:
 @click.option(
     "--no-knowls",
     is_flag=True,
-    help="Produce html-format output without knowls (e.g. for previews of substrees)",
+    help="Supress knowl generation for html-format output (e.g. for faster previews of individual sections)",
 )
 @nice_errors
 def build(
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -405,6 +405,11 @@ def init(refresh: bool, files: List[str]) -> None:
     type=click.STRING,
     help="xml:id of the root of the subtree to be built.",
 )
+@click.option(
+    "--no-knowls",
+    is_flag=True,
+    help="Produce html-format output without knowls (e.g. for previews of substrees)",
+)
 @nice_errors
 def build(
     target_name: str,
@@ -412,6 +417,7 @@ def build(
     generate: bool,
     no_generate: bool,
     xmlid: Optional[str],
+    no_knowls: bool,
 ) -> None:
     """
     Build [TARGET] according to settings specified by project.ptx.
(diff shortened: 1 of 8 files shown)


pretext/cli.py
```diff
@@ -408,7 +408,7 @@ def init(refresh: bool, files: List[str]) -> None:
 @click.option(
     "--no-knowls",
     is_flag=True,
-    help="Produce html-format output without knowls (e.g. for previews of substrees)",
+    help="Supress knowl generation for html-format output (e.g. for faster previews of individual sections)",
```
StevenClontz: This isn't quite right. Knowls are already skipped; this just adjusts things so the page uses hyperlinks instead of knowls.

pretext/cli.py
```diff
@@ -405,13 +405,19 @@ def init(refresh: bool, files: List[str]) -> None:
     type=click.STRING,
     help="xml:id of the root of the subtree to be built.",
 )
+@click.option(
+    "--no-knowls",
+    is_flag=True,
+    help="Suppress knowl generation for html-format output (e.g. for faster previews of individual sections)",
```
oscarlevin, line 411: I reworded the help to make it clear that this is not for keeping content unknowled, but just for previews.

oscarlevin (approved):
If you are happy with my suggestion, I'm good with merging this.

pretext/cli.py
StevenClontz: It's not really "for faster previews of individual sections" though. It's for using hyperlinks rather than knowls (since the knowl files aren't generated either way for subtree biulds).

How about:

```suggestion
    help="Use hyperlinks instead of knowls (e.g. for previewing individual sections when knowl files from other sections may not exist)",
```

oscarlevin: That's great.  And this is good, because clearly I didn't not understand it until now.

StevenClontz: ah, you didn't commit my suggestion though?

This is a feature I've started using a lot, where instead of committing over other folks' PRs, you can use this button:

image

to suggest a change, and the other person can commit the change if they agree.

I'll make a hotfix PR to implement this.

oscarlevin:
oooh.  Maybe I need to stop working for the night!

PR #741, opened 2024-06-16
Include HHMMSS in dev version

StevenClontz:
This change includes HHMMSS in dev/nightly releases.

Rationale: should we want to push an extra dev release between nightlies, this ensures order is maintained. (Inspired by https://github.com/PreTeXtBook/pretext-cli/pull/740 where having a dev release available would have been handy to expriement with Codechat in a vanilla PreTeXt project.)

### DIFF
--- a/scripts/prep_nightly.py
+++ b/scripts/prep_nightly.py
@@ -61,7 +61,7 @@ def main() -> None:
     ):
         if line.startswith("version"):
             version = str(line.split('"')[1])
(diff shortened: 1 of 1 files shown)


oscarlevin:
I have no objection to this, but I would rather not get in the habit of making too many sub-nightly releases.

StevenClontz:
> I have no objection to this, but I would rather not get in the habit of making too many sub-nightly releases.

Agreed!

PR #743, opened 2024-06-17
add support for `deploy` attribute on `<target/>`

StevenClontz:


pretext/project/__init__.py
```diff
@@ -113,9 +113,16 @@ class Target(pxml.BaseXmlModel, tag="target", search_mode=SearchMode.UNORDERED):
...
-    # A path to the subdirectory of your deployment where this target will live.
+
+    # Specify whether this target should be included in a deployment
+    deploy: t.Optional[Path] = pxml.attr(name="deploy", default=None)
+    # A non-default path to the subdirectory of your deployment where this target will live.
     deploy_dir: t.Optional[Path] = pxml.attr(name="deploy-dir", default=None)
 
+    # To support skipping `deploy` but specify `deploy_dir`:
+    def to_deploy(self) -> bool:
+        return self.deploy is not None or self.deploy_dir is not None
```
StevenClontz, line 124:
```suggestion
        return self.deploy is not None or self.deploy.lower() != "no" or self.deploy_dir is not None
```
```suggestion
        return self.deploy is not None or self.deploy_dir is not None
```

PR #746, opened 2024-06-18
work on revamping ghaction

StevenClontz:


StevenClontz:
Toying around here: https://github.com/StevenClontz/fictional-fiesta/actions/

PR #747, opened 2024-06-19
codechat wip

StevenClontz:


oscarlevin:
I assume we are safe to close this given that we are essentially abandoning codechat in favor of the new live preview?

StevenClontz:
Indeed

rbeezer:
I removed some CodeChat setup from the PreTeXt examples just the other day.  Not in a position to quote the PR number right this minute.

PR #750, opened 2024-06-19
Add mermaid support

ascholerChemeketa:
Goes with this PR on PreTeXt: https://github.com/PreTeXtBook/pretext/pull/2178


StevenClontz:
Marking as a draft until accepted as part of core PreTeXt, but thanks for linking these up!

StevenClontz:
This is making me think that part of our test suite should be building the sample article; that way we don't need extra tests for odds and ends like this that have passed Rob's muster already.

rbeezer:
> until accepted as part of core PreTeXt

Now part of **PreTeXt**.

StevenClontz:
Y'know @ascholerChemeketa if you're handling mermaid, and I'm working on [pelican](https://github.com/PreTeXtBook/pretext-cli/pull/745), who's got Sebastian and Flounder?

StevenClontz:
#758 includes a test of the sample article, so assuming there's Mermaid stuff there, let's get that merged to main first, then merge into here for a final integration test before merging.

PR #751, opened 2024-06-20
have deploys depend on tests action

StevenClontz:


StevenClontz:
again just keeping @oscarlevin in the loop on these housekeeping PRs but going ahead with merges, so if we notice anything broken we can revert in the next couple of days

PR #755, opened 2024-06-20
correct docstring for deploy

### LINKED ISSUE #683: specifying target for deploy
If I run `pretext deploy`, part of the terminal log is `Since no target was supplied, we will use "html".` (That being the name of my first target.)

OK, so how should I have supplied the target name? If I try `pretext deploy html` (analogous to `pretext build html`) then I get: `Error: Got unexpected extra argument (html)`.

And if I do `pretext deploy -t html`, I get `Error: No such option: -t`.

`pretext deploy --help` says `Usage: pretext deploy [OPTIONS]` but then says
```
  Automatically deploys most recent build of [TARGET] to GitHub Pages, making
  it available to the general public. Requires that your project is under Git
  version control properly configured with GitHub and GitHub Pages. Deployed
  files will live in the gh-pages branch of your repository.
```
This latter part suggests the usage is meant to be `Usage: pretext deploy [TARGET] [OPTIONS]`.

This is just a small observation right now, but if I had two targets I wanted to deploy (early determinants and late determinants?) then I would need to specify the deploy target somehow.

StevenClontz:
Closes #683

@Alex-Jordan to choose which targets get deployed, set `deploy="yes"` (in current nightlies) or `deploy_dir="whatever"` on those targets. (If nothing is set, then deploy should still deploy the default target, but explicit is better.)

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -842,7 +842,7 @@ def deploy(
     ctx: click.Context, update_source: bool, stage_only: bool, preview: bool
 ) -> None:
     """
-    Automatically deploys most recent build of [TARGET] to GitHub Pages,
+    Automatically deploys project to GitHub Pages,
     making it available to the general public.
     Requires that your project is under Git version control
     properly configured with GitHub and GitHub Pages. Deployed


StevenClontz:
My guess is that tests are failing due to Asymptote server issues (no better reason to finally do what Sean's been asking for and switch CLI to use local asymptote).

StevenClontz:
> My guess is that tests are failing due to Asymptote server issues (no better reason to finally do what Sean's been asking for and switch CLI to use local asymptote).

disregard that, I just suck

oscarlevin (approved):
Nice!

PR #758, opened 2024-06-20
chore: refactor "installing" resources to `~/.ptx`

StevenClontz:


StevenClontz:
Now all static files are installed to the disk at `$HOME / .ptx / PTX_VERSION / core|templates|whatever` in the same manner, making getting at these things much easier using a unified `resources` module.

StevenClontz:
implemented to make https://github.com/PreTeXtBook/pretext-cli/pull/745 less of headache

StevenClontz:
Also we just dump everything from the core pretext repo into that folder. That means we have access to things like the sample article, which I might write a test for...

StevenClontz:
It appears that the sample article uses https://github.com/PreTeXtBook/pretext/blob/ac1ca3ca67c9512059afd6fd37a714a7fc988a5d/pretext/pretext.py#L1034 but we haven't implemented it. @rbeezer which part of the sample article uses this feature?

StevenClontz:
Ah I see this is reverted at ac1ca3ca67c9512059afd6fd37a714a7fc988a5d

StevenClontz:
@dbrianwalton we'll need an update to this repo to get your dynamic exercises working

oscarlevin (approved):
I did not read all of the code carefully, but I mostly agree with what you are trying to do here and trust that you have tested this sufficiently.

My only concern is that if we keep multiple versions of the .ptx  resources (for different versions), then we will have lots of copies of the sample article.  Just a little worried about filesize.

Okay, also one more thing: there should be an easy way to refresh the resources in case they get corrupted some how.

StevenClontz:
> My only concern is that if we keep multiple versions of the .ptx resources (for different versions), then we will have lots of copies of the sample article. Just a little worried about filesize.

> Okay, also one more thing: there should be an easy way to refresh the resources in case they get corrupted some how.

`rm -rf ~/.ptx` is a fix for both. I thought about adding a command like `pretext clear cache` but I don't think it's super important as folks don't upgrade *that* often (and if they do that's not a hard command).

StevenClontz:
An issue with packaging this: the relocated zip files aren't being grabbed...

PR #764, opened 2024-06-21
Use oscarlevin:full by default in codespace

StevenClontz:
Would have prevented https://groups.google.com/g/pretext-support/c/5EhXaz_MHwA/m/iRtDPVbRCAAJ - I don't think the extra minute saved is worth the headache of explaining to users how to get Sage installed in their codespace.

### DIFF
--- a/templates/.devcontainer.json
+++ b/templates/.devcontainer.json
@@ -15,9 +15,9 @@
   "name": "PreTeXt-Codespaces",
 
   // This Docker image includes some LaTeX support, but is still not to large.  Note that if you keep your codespace running, it will use up your GitHub free storage quota.  Additional options are listed below.
-  "image": "oscarlevin/pretext:small",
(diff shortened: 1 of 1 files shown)


oscarlevin:
I'd like to think about this some more.  It is less about the extra startup time than it is about the storage limit imposed by codespaces (and the new scrutiny that github is apparently putting on edu accounts for the bonus).  

What about adding better error messages or even automatic switches if a project fails to build because of a sageplot?

StevenClontz:
I've had a couple instances where users hit the limit, but not very many. I'd prefer configuring the retention period for codespaces to be shorter (I do this personally), but I don't think that can be configured in `.devcontainer`, so it's still an extra step for someone to make.

oscarlevin:
Okay, I think you are right.  We will try it this way for now.

PR #765, opened 2024-06-21
use utils.exit_command to exit with errors and remove errorhandler dep

### LINKED ISSUE #761: use utils.exit_command to exit with errors and remove errorhandler dep

StevenClontz:
Closes #761


oscarlevin:
I think we might want to re-backtrack on this.  Turns out we cannot call exit commands inside `at_exit`.  Doing so causes an exception and traceback to be printed.

Issue #772, opened 2024-06-27
new core resources vs symlink core issues


oscarlevin:
I believe starting with #758, there appears to be some issue with core resources being installed in `~/.ptx` and symlink core.

- I don't think that undoing symlink core by running `fetch_core.py` works anymore.
- At least on Windows, the symlinking is not working.  I don't know if this was a problem before with just how windows manages symbolic links.
- Previously we checked the core commit sha to decide whether we needed to update the core resources.  It doesn't seem like we are doing this any more, opting instead just to look for folders with the version name installed.  This might be why the first bullet is happening.
- At least while trying to develop, I had a failure to install all the core resources (I think because I hadn't fetched core with poetry first) and did not get any indication that something was wrong.  I think that the previous messages telling the user that new static resources were being installed were helpful, but they are gone now.

Thoughts @StevenClontz ?

oscarlevin:
More precisely, the first time I run `pretext --version` inside poetry, I got an error, that it couldn't find `resources/templates.zip`.  Okay, so I forgot to zip templates.  But the second time I ran it, this error didn't happen since it saw the existence of 2.4.2.  So we should probably remove the resource directory if it isn't created correctly.  Or only copy it to the user's directory if it was successfully created.

oscarlevin:
Okay, with #774, `symlink_core.py` correctly creates the dev environment.  However, it does not get unlinked ever (unless you delete the latest `.ptx` folder).  We should make it easier for developers, especially of core pretext.  Some ideas:

- [ ] Create a `dev_setup.py` script that fetches core, bundles resources, and symlinks core.
- [ ] Create a script or option on the `dev_setup.py` script that unlinks core resources.
- [ ] Explore whether it is possible to do this somewhat automatically with poetry
- [ ] Update the documentation
- [ ] Post announcement on pretext-dev

StevenClontz:
I'd add:

- [ ] Ensure the script runs upon creation of a pretext-cli devcontainer

oscarlevin:
Agreed.  

A further weirdness: After doing a symlink of core, the `core/pretext.py` file is a symbolic link.  Then upon a further `fetch_core.py` run, it actual core pretext.py file that gets updated in my pretext repository.  No biggie, it just whipped out my work twice now.

PR #774, opened 2024-06-27
Improve core dev tools

oscarlevin:
The first commit fixes the `symlink_core.py` issue mentioned in #772.  Will add some more features to this soon.


oscarlevin:
I'm going to merge this for now, so at least `symlink_core` works.  I'll think about a more robust development environment setup later.

PR #775, opened 2024-06-28
Dev environment improvments

### LINKED ISSUE #772: new core resources vs symlink core issues

oscarlevin:
This will close #772.  Note the the request to make it easy to use in a codespace isn't really addressed, since the directions have basically not changed (except there is now a unlink_core script).


oscarlevin:
@StevenClontz, I'm going to merge this so I can start using it, but when you get a chance, please take a look to check that I haven't done anything too crazy.

scripts/symlink_core.py
```diff
@@ -1,18 +1,24 @@
...
-    else:
-        link_path.unlink()
+    # Remove the current link or directory
+    utils.remove_path(link_path)
+    # Create a symlink to the core directory
     link_path.symlink_to(core_path)
+
+    # Remove the current pretext/core/pretext.py file
+    script_link_path = Path("pretext").resolve() / "core" / "pretext.py"
+    script_core_path = core_path / "pretext" / "core" / "pretext.py"
```
StevenClontz, line 18: should this be `core_path / "pretext" / "pretext.py"`?

StevenClontz: @oscarlevin

PR #776, opened 2024-06-28
fix incorrect path

oscarlevin:
This corrects an error in #775


scripts/symlink_core.py
```diff
@@ -15,7 +15,7 @@ def main(core_path: Path = Path("../pretext")) -> None:
 
     # Remove the current pretext/core/pretext.py file
     script_link_path = Path("pretext").resolve() / "core" / "pretext.py"
-    script_core_path = core_path / "pretext" / "core" / "pretext.py"
+    script_core_path = core_path / "pretext" / "pretext.py"
```
StevenClontz, line 18: ah. :-)

PR #778, opened 2024-07-08
Add support for dynamic-substitutions (dynamic-subs)

dbrianwalton:
* New generated asset: dynamic-subs

---

I ran the formatting and attempted the tests but didn't have sage installed locally to finish the test. I did verify that the generated assets file was created when called and that the print target successfully generated on its own.


oscarlevin:
This looks great.  Thanks!

Issue #780, opened 2024-07-10
`at_exit` sys.exit conflict


oscarlevin:
The new `sys.exit(1)` call in the `at_exit()` cause a stack trace.  So it works as intended in a CI run, but not great for users.  @StevenClontz, thoughts about going back to your previous attempt at this? (In case you want something to do later this week).

oscarlevin:
I believe the only time we really need to worry about having build errors for automatic testing is on a `build`.  So we should just run a check for the memory handler at the end of the build, and `sys.exit(1)` there.  We will use the `at_exit` just to report that everything completed successfully (as it was intended).

PR #781, opened 2024-07-11
move exit command to end of appropriate cli commands

oscarlevin:
To address #780.  This does what we want, without importing `at_exit`.  But slightly more brittle code, requiring new CLI commands to remember to call `utils.exit_command` as their last thing.


StevenClontz (commented):
See proposal at #782 (targeting this branch, not main)

StevenClontz (approved):

PR #782, opened 2024-07-11
use main result callback, defer systemexit to exit_command

StevenClontz:
I think this is a bit cleaner - just always check for errors at the end of a CLI run, and not have to worry about a bunch of finallys.


oscarlevin:
This is lovely!  Thanks!!

PR #786, opened 2024-07-15
make pelican an extra

bnmnetp:
This makes pelican an "extra"

A `pip install pretext` will not include pelican.  However `pip install pretext[homepage]` will. 

In `project/__init__.py` if pelican does not import then we set `USEPELICAN` to False.  Later if we do hit the case where the strategy is "pelican_default" or "pelican_custom" we check the `USEPELICAN` flag and emit an error saying that pelican needs to be installed.

If the flag is false then we won't ever call the function from utils that sets up the pelican settings so we should only need to catch the import error there.


oscarlevin (changes requested):
This looks great.  Would you mind adding a line to the changelog under "###fixes" to say that for compatibility, to get the landing page features, you must install pretext with `pip install pretext[homepage]`?  Then I can merge and release the patch.

StevenClontz:
Our test suite needs to be updated to not assume pelican is always available. It's also not clear to me what a deployment should look like when pelican is not available.

bnmnetp:
OK, at a minimum I updated the workflow to ensure that pelican is installed for the tests.  You can see that in the last one it did not install pelican by default, and there was the log message about installing pelican.

After further testing on my own I learned that I needed to rebuild poetry.lock in order for the extra part to actually work.

oscarlevin (approved):

PR #789, opened 2024-07-16
Build Heuristics

### LINKED ISSUE #788: Guess target if target name is incorrect
If a user types `pretext build html` and there is no `html` target (only `web`) and only one `html`, why not just build html and warn the user?

siefkenj:
When a build target is not found, the build format is checked as well. If one is matching, a warning is printed and that target is built.

This means a user can mistakenly do
```
pretext build pdf
```
when the actual build target is `print`, and it will still work.

Closes #788

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -1326,6 +1326,15 @@ class Project(pxml.BaseXmlModel, tag="project", search_mode=SearchMode.UNORDERED
             # return first target matching the provided name
             return next(t for t in self.targets if t.name == name)
         except StopIteration:
+            # if there is no target matching, check if there is a format
+            # that matches. Use that as a backup with a warning
+            for target in self.targets:
+                if target.format == name:
+                    true_name = target.name
+                    log.warning(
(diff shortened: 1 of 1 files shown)


StevenClontz:
Closes #788

pretext/project/__init__.py
```diff
@@ -1326,6 +1326,15 @@ def _get_target(
             # return first target matching the provided name
             return next(t for t in self.targets if t.name == name)
         except StopIteration:
+            # if there is no target matching, check if there is a format
+            # that matches. Use that as a backup with a warning
+            for target in self.targets:
+                if target.format == name:
```
StevenClontz, line 1332: I have a slight preference to ensure `name` is a valid `format` here, but it's not a strong preference.

siefkenj: Shouldn't `format` be a valid format unless the person's `project.ptx` is not schema-compliant?

StevenClontz: yeah, I think that is sufficient guardrails now that you mention it.

StevenClontz:
Can we add a test for this in our suite?

oscarlevin:
Thanks!

PR #791, opened 2024-07-17
fix duplicate id bug

oscarlevin:
Codechat parses the xml it complains and causes a critical error when there are duplicate `xml:ids` even if they are in separate components only one of which is included in a particular `version`.  

This just wraps the codechat `map_path_to_xml_id` in a `try/except` as a stop-gap until this is fixed.  Or maybe we never fix it, since it is very unlikely someone with this sort of structure would use codechat anyway.

FYI @StevenClontz and @bjones1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ Instructions: Add a subsection under `[UNRELEASED]` for additions, fixes, change
 
 ## [UNRELEASED]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


bjones1:
@oscarlevin, do you have some PreTeXt source that reproduces this bug? If so, I'll be happy to provide a fix that avoids the exception.

oscarlevin:
To reproduce, put the same xml:I'd on two elements, and put different `@component` on them.  In the publisher file, add a version include just one of them.

PR #807, opened 2024-07-30
Development Environment Improvements

oscarlevin:
This removes an often missed step of running `scripts/bundle_resource.py` (it is not part of the `fetch_core.py` script).  Readme is also cleaned up to make instructions easier to follow.


oscarlevin:
FYI @StevenClontz

PR #810, opened 2024-08-01
add support for site.json instead of site.ptx

StevenClontz:
Long-term, I'd like `site.ptx` to have full functionality, but since I need this *today*, I'm adding stopgap support for JSON files (so we can assign more than just string values to site configuration).


StevenClontz:
(i'm sneaking out a nightly release from this branch so I can use this at https://github.com/TeamBasedInquiryLearning/library so no rush to review)

oscarlevin (approved):
This seems reasonable.

PR #828, opened 2024-09-06
fix up github action

### LINKED ISSUE #824: GitHub Action doesn't work for older pretext versions
The current push action assumes a recent version of the CLI in that it calls `pretext build --deploys`.  But before that, the action installs the pretext version specified in `requirements.txt`.  This causes a failure not because the book is broken but because `requirements.txt` is out of date.

oscarlevin:
The action for push will now check to ensure that version 2.5 or greater is installed before running `pretext build --deploys`, and run `pretext build` if not.  

Closes #824

### DIFF
--- a/templates/pretext-cli.yml
+++ b/templates/pretext-cli.yml
@@ -24,8 +24,21 @@ jobs:
             - name: install deps
               run: pip install -r requirements.txt
 
+            - name: install local ptx files
+              run: pretext --version
+
             - name: build deploy targets
-              run: pretext build --deploys
+              run: |
+                version="$(pretext --version)"
+                major="$(echo $version | cut -d '.' -f 1)"
+                minor="$(echo $version | cut -d '.' -f 2)"
+                if [ "$major" -ge 2 -a "$minor" -ge 5 ]; then
+                    echo "PreTeXt version is 2.5 or greater; using new build command"
+                    pretext build --deploys
+                else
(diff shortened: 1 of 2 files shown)


templates/pretext-cli.yml
```diff
@@ -24,8 +24,21 @@ jobs:
...
+                version="$(pretext --version)"
+                major="$(echo $version | cut -d '.' -f 1)"
+                minor="$(echo $version | cut -d '.' -f 2)"
+                if [ "$major" -ge 2 -a "$minor" -ge 5 ]; then
+                    echo "PreTeXt version is 2.5 or greater; using new build command"
+                    pretext build --deploys
+                else
+                    echo "PreTeXt version is less than 2.5, using old build command"
+                    pretext build
+                fi
```
StevenClontz, line 41: Is there a clever way we can configure `pretext-cli.yml` to know the version of PreTeXt that generated it? I'd just as soon reject any attempt to use CI/CD with a mismatched PreTeXt version and GitHub action.

StevenClontz (approved):
I think this is non-harmful, at the least.

templates/pretext-cli.yml
oscarlevin: If a user generates the `pretext-cli.yml` with 2.6 but decides to keep 2.4 in `requirements.txt` then it would still fail.

PR #860, opened 2024-11-08
Update frontmatter.ptx for bibinfo

Tanaquil18:
started getting warnings when I built and found an example to follow in the sample book.


oscarlevin:
Thanks for this.  Very helpful contribution.

PR #878, opened 2024-12-17
Reworking view

skiadas:


pretext/server.py
skiadas, line 1: Interesting that this didn't appear to be a problem when I was testing locally. I could have sworn I had tested with more than one entry. 
Would this be better done by simply adding the newline in the toFileLine function instead?

oscarlevin: Yes, I like that better actually.  I'll make that change.

oscarlevin:
I removed a hardcoded port and now the tests are passing.  I think though that it would be helpful to add a check to see if the current server is actually still up and working.  I ran into some issues where some entries in the running_servers file did not get removed but the server was killed (say a power outage).  I'll poke around a bit more before merging.

skiadas:
Hm in theory it's supposed to go through and check them with the isActiveServer method. But maybe the timing of when that happens isn't right? At a quick glance maybe `get_running_servers` needs to filter the entries based on isActiveServer. I think right now the cleanup attempt only happens in add_server_entry which is probably too late.

Possibly the least code-changing solution would be to add a conditional check in the call to "next" in `active_server_for_path_hash`, to limit it to those `info` who also respond True to `info.isActiveServer()`.

oscarlevin:
Thanks for that hint, it helped me track down what was happening.  I think its working now, but I want to make sure that codespaces work as expected before merging and releasing a new version.

oscarlevin:
Thanks for this.  Very nice improvement.  This doesn't have any effect on the codespaces.  It would probably be worth implementing something similar there, but that's a task for later.

pretext/server.py
```diff
@@ -0,0 +1,201 @@
...
+class RunningServerInfo:
+    """A simple dataclass to hold the information in the running servers file."""
+
+    pathHash: str
+    pid: int
+    port: int
+    binding: str
+
+    @staticmethod
+    def fromFileLine(line: str) -> RunningServerInfo:
```
StevenClontz, line 27: Note that methods in Python should be `snake_case`: https://peps.python.org/pep-0008/#method-names-and-instance-variables

skiadas: It's funny that the logging module doesn't follow that convention.  I've definitely been bad at following the convention, as my mind really doesn't like it at all.
Would it perhaps be possible to add this as something reported by the styling tools?   https://github.com/PyCQA/pep8-naming

StevenClontz: I think Java mixes `snake_case` and `camelCase`, if I recall? I know it's a convention elsewhere, but I try to stick with pep8 for Python, even if Python itself violates it in some core libraries...

And yeah, really this should be caught by our linter.

skiadas:
@oscarlevin Regarding to the recent email to pretext-support, I think there is currently a bug if the process terminates abnormally. To demonstrate:

- Start with `project view`
- From another shell, `kill -9 ...` the process created by project view.
- Back in the shell that ran the `project view`, try to run it again

Solution:

- In [is_active_server](https://github.com/PreTeXtBook/pretext-cli/blob/a4a603a03102a58785bb29755f86b035df5d3c3e/pretext/server.py#L43) we must surround the call to `psutil.Process` in a try-catch, and jus return False in the catch part.

I am currently having some problems running a dev environment or I would make a pull request. I'll try to get something sorted.

PR #881, opened 2024-12-18
Enforce pep8-naming conventions in tooling

skiadas:


pretext/server.py
```diff
@@ -13,27 +13,32 @@
 # Get access to logger
 log = logging.getLogger("ptxlogger")
 
+# Limit for how many entries to allow in the server file
+# before attempting to clean up non-running entries.
+# Note: This is not a limit to the number of concurrent servers.
+PURGE_LIMIT = 10
+
```
StevenClontz, line 20: This appears to be unrelated to the pep8 stuff, so don't forget to open a new PR to implement this

skiadas: It was a constant that was defined inside the one function that uses it, but pep8-naming complained about the all-caps use there, so I moved it to the top. If I leave it back without renaming we would be getting pep8-naming errors.

StevenClontz: nm, I see this was moved out to be a constant.

PR #886, opened 2024-12-25
Fix view for the case where a process is terminated abnormally

skiadas:
If a process no longer exists, `psutil.Process` throws an exception and we must catch it.


oscarlevin:
Thanks for the quick holiday fix.

PR #903, opened 2025-01-16
Require node version check for cssbuilder npm install

ascholerChemeketa:
In https://github.com/PreTeXtBook/pretext/pull/2371, a minimum node version (14) is being listed in the cssbuilder package.json. npm won't actually obey that unless invoked with `--engine-strict=true`.

If there is no engine requirement in the package.json, the flag does nothing, so there shouldn't be any issues with using it with an old version of PreTeXt.


oscarlevin:
Thanks!

PR #911, opened 2025-02-04
add temporary reload of core to avoid cached publication vars for rs

oscarlevin:
This will help rs until we can fix the issue with with core storing a version of the publication variables.

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -8,6 +8,7 @@ import shutil
 import tempfile
 from pathlib import Path
 from functools import partial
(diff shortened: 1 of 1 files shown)


oscarlevin:
@bnmnetp, I've merged this and just done an extra "nightly" release.  If you want to take that for a spin (including testing the new build caching), and everything works well, I can do a regular release later today.

bnmnetp:
Thanks @oscarlevin I've got a couple small things to finish up but should be able to get to this today.  Thanks for the extra release!

PR #912, opened 2025-02-04
remove cached variables at the start of the build

bnmnetp:
So, the reload trick did not work.   But this solution does.  

I just built welcomecs2 and dmoi-4 back to back and they both have the right theme!

Tagging: @ascholerChemeketa 

This still feels like a hack until we get the right solution.  But reasonable people might disagree on that.

### DIFF
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -8,7 +8,6 @@ import shutil
 import tempfile
 from pathlib import Path
 from functools import partial
-from importlib import reload
 
 from lxml import etree as ET  # noqa: N812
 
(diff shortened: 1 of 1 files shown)


bnmnetp:
Also @oscarlevin - the new cache stuff looks like it is fine.

ascholerChemeketa:
That seems as good a hack as any. 

Still think the better long term answer is to pull the caching out of pretext and add some hooks to pass in an already parsed dict of pub vars. If the state causes problems, let's get rid of it.

oscarlevin:
Ah, that's gorgeous.  Very nice.  I'll merge this and release 2.13 later today.

PR #923, opened 2025-02-11
allow deployment of non-clean project

oscarlevin:
This changes the behavior of a `pretext deploy` to still deploy even if there are uncommitted changes to the repository.  Seems to work just fine.  Changed the errors to warnings.

Thanks for the suggestion, @davidaustinm .

@StevenClontz, see any issue with this?  I think the original code in the deploy function was written by you.


StevenClontz:
I don't see a wise use-case for deploying output that's unlikely to match its source, but maybe I'm overlooking something? That was the rationale anyway.

StevenClontz:
(Of course, these days I believe builds+deploys should be containerized in the cloud, and all authors should do is push source, not their local machine builds.)

oscarlevin:
The use case is an instructor who makes a few changes before class, builds and views, and then deploys on the way to class, without wanting to think about a commit message (or perhaps they are in the middle of what will be a larger commit)

davidaustinm:
It has happened a few times this semester that I want to post notes for students a few minutes before class.  I have pushed everything that I want in my repository, but maybe there's an untracked file, such as a set of notes, or even crude PTX source, for what is to come next.  I don't want that to be part of the repo, but I need to deploy right away and I'm not able to without some maintenance of the repository.  It seems like I should just be able to get my output up for others to view and then worry about the state of the repository at another time.

StevenClontz:
Thanks. I'm okay with this change.

The scare-quotes-"correct" way to support this might be to support branch builds/deploys. I have a proof of concept of this at e.g. https://github.com/TeamBasedInquiryLearning/library/pull/592#issuecomment-2646013040 - as an author hacks on a branch, GitHub automatically builds and ships a preview of that change. I've used this to write an activity, try it in the classroom, then make edits before merging into main.

But that implementation requires Cloudflare Pages, which is free, but is another Thing not built into GitHub.

PR #946, opened 2025-03-23
added prefigure as an asset for revealjs format

davidaustinm:
This change enables PreFigure diagrams to be included in slideshows with the revealjs format.


oscarlevin:
Thanks!

PR #952, opened 2025-04-07
add no-push option to pretext deploy

oscarlevin:
This also adds some debug info in case of failures

### DIFF
--- a/pretext/cli.py
+++ b/pretext/cli.py
@@ -945,11 +945,12 @@ def view(
 )
 @nice_errors
 @click.pass_context
-@click.option("-u", "--update-source", is_flag=True, required=False)
-@click.option("-s", "--stage-only", is_flag=True, required=False)
-@click.option("-p", "--preview", is_flag=True, required=False)
+@click.option("-u", "--update-source", is_flag=True, required=False, help="Commit and push changes tot he source files at the same time as deploying output.")
+@click.option("-s", "--stage-only", is_flag=True, required=False, help="Create a staged deployment, but do not deploy.")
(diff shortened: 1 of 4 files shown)


pretext/cli.py
```diff
@@ -945,11 +945,39 @@ def callback(actual_port: int) -> None:
...
+    "--preview",
+    is_flag=True,
+    required=False,
+    help="Preview the staged deployment, but do not actually deploy.",
+)
+@click.option(
+    "--no-push",
+    is_flag=True,
+    required=False,
+    help="Do not push to remote.  Useful for CI/CD workflows or in case of authentication errors.",
```
StevenClontz, line 973: It's not clear to me what this does.

StevenClontz: Does it only push the `gh-pages` branch to deploy? (Not the main/active branch?) Or is this supposed to trigger a CI/CD workflow?

oscarlevin: It does not push anything.  So the user (or action) needs to run `git push origin gh-pages` manually.

StevenClontz: Action deploys should be using https://github.com/marketplace/actions/deploy-github-pages-site rather than a gh-pages branch, so the CI/CD note here is a little confusing.

oscarlevin: Define "should".  I would like to allow a workflow where an author could deploy from the command line or an action, which I think needs to use the pretext deploy command.  But that doesn't seem to work on an action for authentication reasons.  I'm hoping that this would let the action deploy the branch itself

PR #966, opened 2025-04-17
Update installLatex.sh

StevenClontz:
Testing now but I think this is necessary to build latex-images (it definitely was for my project).


StevenClontz:
Okay, I only need this because (of course) my project includes

```xml
    <latex-image-preamble>
\usepackage{tikz}
\usepackage{tikz-cd}
    </latex-image-preamble>
```

but I think tikz-cd is pedestrian enough that it's good to include in our defaults.

oscarlevin (approved):
Great, thanks.

oscarlevin:
We should add any package that any author needs.  So I expect this will update frequently.

PR #972, opened 2025-05-10
Set up cache asset generation for prefigure

oscarlevin:
This won't work until PR 2500 on core gets merged, but then it should be good to go.

### DIFF
--- a/pretext/__init__.py
+++ b/pretext/__init__.py
@@ -18,7 +18,7 @@ from single_version import get_version
 
 VERSION = get_version("pretext", Path(__file__).parent.parent)
 
-CORE_COMMIT = "bc9b29b568caa84ba1a099b1bbf3125167ac241c"
+CORE_COMMIT = "9be1596d6ecab588c8a7c1a5c437c7785b895f23"
 
 
 def activate() -> None:
--- a/pretext/project/__init__.py
+++ b/pretext/project/__init__.py
@@ -1054,6 +1054,11 @@ class Target(pxml.BaseXmlModel, tag="target", search_mode=SearchMode.UNORDERED):
                         xmlid_root=xmlid,
                         dest_dir=self.generated_dir_abspath() / "prefigure",
                         outformat=outformat,
+                        ext_converter=partial(
+                            generate.individual_prefigure,
+                            cache_dir=self.generated_cache_abspath(),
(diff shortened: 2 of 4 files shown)


rbeezer:
https://github.com/PreTeXtBook/pretext/pull/2500 is in

mitchkeller:
This needs a fix before merging. `individual_prefigure`'s argument `tmp_dir` isn't coming in a path, so `output_file = tmp_dir / asset_file.with_suffix(f".{format}").name` is throwing an error because both operands are strings.

oscarlevin:
Thanks @mitchkeller, I think I got that part fixed.

But two questions for you (maybe) and (definitely) @davidaustinm: When I generate prefigure images, I sometimes get an extra xml file with `-annotations` appended to the filename.  Any thoughts on what this should do for the caching?  Are those created every time... which format is built? Are the built reliably?  That is, if I know the svg for that image has been built and cached, can I assume that not seeing the `-annotated` version would mean that there isn't one?

The other question: when I tried to build with `--all-formats` I would get strange errors.  The first time I tried to build, a particular svg would be reported as not existing.  But the second time, it was a new svg that didn't exist.  Could this be because I changed the order things were generated in core?  

Happy to talk about this in person, but wanted to get a record of what I'm running into so I don't forget.

davidaustinm:
The *-annotations.xml file is only created when an author includes #annotations in their PreFigure source and then builds to the svg format.  Ideally, they would also be cached;  if an author modifies the annotations, but not the graphical part of the PreFigure source, they would still want that diagram to be recompiled.  Similarly, an author could modify their PF source leaving the graphical part alone and adding annotations that were not previously there.  Again, they would want to rebuild.

We can talk about the errors you're seeing, but I suspect it is the order in which things are built.  An svg is always created, and that is used to build pdf, and png.  I've forgotten how the script was originally ordered, but I suspect it took advantage of that and just built the svg first for the "svg" format and then used that to build to the other formats.  Let's talk tomorrow.

davidaustinm:
In the core pretext individual_prefigure_conversion function, we should build to the "svg" format last.  Building to pdf and png remove the intermediate svg that is produced.  We are currently building svg, then pdf, then png.  The svg that is created initially is wiped out by the other builds.

Issue #1000, opened 2025-06-18
HTML syntax is error messages is ineffective


rbeezer:
Mis-reported, for reasons that will be obvious:

https://github.com/RunestoneInteractive/rs/issues/761

Short version: CLI failure messages are being shown in a webpage, by Runestone, and in this case `<document-id>` went missing.

Might be more extensive than this one case?

PR #1005, opened 2025-06-20
remove '<' characters from log messages

oscarlevin:


oscarlevin:
This will close #1000

Issue #1026, opened 2025-07-12
Workflow builds failing due to missing Node.js


daverosoff:
Since updating to 2.23.0, I am not able to pass the automated build tests when pushing.

Expected behavior: node is installed in the pretext-cli workflow and my CSS/Prefigure compiles
Actual behavior: many errors due to missing node

I'd be happy to take a crack at this and submit a PR if such contribution is welcome.

oscarlevin:
It would be great if you could try to figure this out.

This is because the new dynamic substitutions require mode 22.10 or higher.  I have tried to install this in the docker image, but it's not working for CI builds (I think I did get it working for dev containers).  

One approach would be to add a node activate step in the action.  Look at the test action for the CLI; that seems to work.

oscarlevin:
@bnmnetp, that looks cool.  I was trying to use NVM which was a pain for CI (and the nodejs that comes with debian is out of date)

PR #1027, opened 2025-07-14
Added node setup as in tests.yml

daverosoff:
This seemed to do the trick fixing #1026.


oscarlevin:
Excellent.  Thanks for this.

PR #1072, opened 2025-09-15
format frontmatter in article template

StevenClontz:


oscarlevin (approved):
This is a good change.  I think we need to actually add more to the template (add a titlepage element).  Would you like to add those here?  I am happy to merge this as is if you don't think you can add to it soon.

PR #1083, opened 2025-11-15
Add exception handling for is_active_server. Close #1082

### LINKED ISSUE #1082: Logic for recognizing active viewing server needs to handle errors in connection search
The code in lines https://github.com/PreTeXtBook/pretext-cli/blob/121246df188f0878abf5e431056fafd6491a0079/pretext/server.py#L56-L78 has the possibility of failing if  the pretext process doesn't have sufficient permissions to call `p.net_connections("all")`. In some systems a `psutil.AccessDenied` error may be raised, which is currently not being caught and handled.

I believe the solution is to simply place that entire block within the `try` block on line 51, and add an `except` clause for `AccessDenied`.  

@oscarlevin would you like me to prepare a pull request?

I think it is an interesting question why this isn't a problem on my local mac system. I'm not sure how to troubleshoot it more precisely.

skiadas:


oscarlevin:
Thanks, this should be much better now.

PR #1097, opened 2025-12-10
update lxml dependency

oscarlevin:
To work with windows and python 3.14, we should use lxml v6.x.


oscarlevin:
Very strangely, it seems like using lxml v6 breaks having path names with spaces in them.  I don't understand why.  I'll hold off this fix until I can get that sorted out.

PR #1158, opened 2026-05-28
Support xinclude pub files

StevenClontz:


oscarlevin:
This looks reasonable to me.  I don't know how useful modular publication files will be, but this shouldn't hurt

PR #1201, opened 2026-08-01
Validation and external directory updates

oscarlevin:
This switches the `pretext validate` command to use core pretext's version.  Also included is an experimental "jing-free" validation tweak.

It also fixes issues with logging and updates templates to use the new specification of external directories in docinfo.


oscarlevin:
Waiting on a merge from core, which should fix all the tests

